Issue #41, opened 2026-07-01
[Backend Migration 09] Implement password auth and JWT sessions


GonzaloSecades:
Parent spec: #18.
Related epic: #22.
Spec source: `docs/specs/18-go-backend-replacement/spec.md`.

Graphify note:
- Use the current graph output to confirm impacted endpoints, flows, and business logic before implementing changes.
- Refresh the graphify artifacts before handing this ticket in or before starting another migration ticket.

Implement core owned authentication in the Go backend.

Acceptance criteria:
- Register creates a user with a password hash only.
- Login accepts verified email/password and returns a short-lived JWT access token.
- Access-token lifetime is configurable and defaults to 30 minutes for development.
- Refresh token is set in an HttpOnly cookie.
- Refresh rotates or renews access from the refresh cookie.
- Logout revokes refresh/session state and clears the cookie.
- Auth-sensitive identity comes from verified token/session state, not request body user IDs.

Verification:
- `cd backend && go test ./...`
- Tests cover register, login, refresh, and logout.

GonzaloSecades:
## Design decisions for this ticket (recorded 2026-07-10, during #38 schema work)

Agreed between Gonzalo and the orchestrator while designing the auth base schema in #38. The #38 tables (`users`, `email_verification_tokens`, `password_reset_tokens`, `refresh_tokens`) already carry the columns these behaviors need (`created_at`, `expires_at`, `used_at`, `revoked_at`, `user_id` indexes, unique `token_hash`).

**IDs:** UUIDv7, generated app-side (`uuid.NewV7()` — native `uuidv7()` needs PG18, we run 17). The DB's `gen_random_uuid()` default is only an ad-hoc-insert safety net.

**Password hashing:** Argon2id, stored as an opaque PHC-encoded string in `users.password_hash`. Raw passwords exist only in transit (TLS assumed/required) and in memory during hashing.

**Token handling:**
- Tokens sent by email are random 256-bit values; only their SHA-256 hash is stored (`token_hash`). Lookup by unique-index hash equality is fine — timing attacks aren't practical against a 256-bit random token.
- **Invalidate prior tokens on issue:** generating a new reset/verification token marks all previous unused ones for that user as used (no multiple live tokens in an inbox).
- **Invalidate reset tokens on password change** through any path.
- **Password reset confirm is one step** (token + new password together), and a successful reset **revokes all active refresh sessions** for the user.
[...]

Issue #27, opened 2026-06-26
[Backend Migration 21] Remove legacy Hono/Drizzle/Neon/Clerk backend after Go parity


GonzaloSecades:
Parent spec: #18.

Graphify note:
- Use the current graph output to confirm impacted endpoints, flows, and business logic before implementing changes.
- Refresh the graphify artifacts before handing this ticket in or before starting another migration ticket.

Remove legacy Hono, Drizzle, Neon, Clerk, and Clerk UI only after the Go backend, generated OpenAPI client, custom auth, and frontend integration are usable. This issue remains blocked until parity is proven.

Acceptance criteria:
- `app/api/[[...route]]` Hono backend code is removed or replaced only by the agreed Next rewrite/proxy configuration.
- Drizzle schema/routes are preserved under `deprecated/legacy` only if still useful as reference.
- Hono, Drizzle, Neon, Clerk, and related packages/scripts/env vars are removed when unused.
- Auth pages/components no longer depend on Clerk.
- `.env.example` documents the new backend, database, JWT, and mail settings.
- Full verification passes after removal.

Verification:
- `bun run lint`
- `bun run build`
- `cd backend && go test ./...`
- backend migration command
- local smoke test for auth and dashboard flows

GonzaloSecades:
Migration dependencies #50 and #51 are now complete and merged. The frontend uses the generated authenticated client for all active domains, owned Go auth is live in the app, and `/api/*` targets Go by default. #27 is now the next migration ticket for legacy Hono/Drizzle/Neon/Clerk teardown after the final parity confirmation. This remains a high-risk human-reviewed step, so I am not adding `agent:ready` automatically.

GonzaloSecades:
Closing. All three children are merged and closed: **#84** (Hono surface),
**#85** (Drizzle/Neon/Clerk), **#90** (parity harness and `pg`), plus **#101**
which fixed the last piece of active guidance still describing the removed
stack. Master is at `638f816`.

## Acceptance criteria, walked one at a time

**1. `app/api/[[...route]]` removed, replaced only by the agreed rewrite/proxy.**
Done in #84. The rewrite is now unconditional — `USE_GO_API` selected nothing
once there was no second backend — and stays in `beforeFiles` as a deliberate
choice rather than a forced one: with the catch-all gone, `afterFiles` would
fire too, but a future Next route under `app/api/` would silently shadow the
backend from that phase. Verified behaviorally, not by reasoning: `/api/health`
[...]

Issue #43, opened 2026-07-01
[Backend Migration 10] Add Go auth middleware and DB RLS session binding


GonzaloSecades:
Parent spec: #18.
Related epics: #21 and #22.
Spec source: `docs/specs/18-go-backend-replacement/spec.md`.

Graphify note:
- Use the current graph output to confirm impacted endpoints, flows, and business logic before implementing changes.
- Refresh the graphify artifacts before handing this ticket in or before starting another migration ticket.

Connect authenticated requests to the Go route layer and PostgreSQL RLS model.

Acceptance criteria:
- Middleware validates Bearer JWT access tokens.
- Middleware exposes the authenticated internal user UUID to handlers.
- DB access sets the authenticated user UUID in the database session or transaction before owned queries run.
- Missing, malformed, expired, or invalid tokens return 401.
- Handlers do not accept user identity from request bodies.
- Cross-user access is blocked by app checks and RLS backstop.

Verification:
- `cd backend && go test ./...`
- Tests cover unauthorized requests and cross-user isolation.

GonzaloSecades:
# Briefing — #43 [Backend Migration 10] Go auth middleware + DB RLS session binding

risk:high. Attended. Human merge review. No unattended runs.

## 1. Ticket

**#43 — Add Go auth middleware and DB RLS session binding**

Acceptance criteria (verbatim):
- Middleware validates Bearer JWT access tokens.
- Middleware exposes the authenticated internal user UUID to handlers.
- DB access sets the authenticated user UUID in the database session or
  transaction before owned queries run.
- Missing, malformed, expired, or invalid tokens return 401.
- Handlers do not accept user identity from request bodies.
- Cross-user access is blocked by app checks and RLS backstop.

Verification: `cd backend && go test ./...`; tests cover unauthorized
requests and cross-user isolation.

## 2. Why this ticket is load-bearing

The #39 RLS policies have existed since that ticket, but **nothing has ever
set `app.user_id` at request time**, so no policy has ever engaged on a real
request. CI only started genuinely enforcing RLS as of the #53-era workflow
fix. Treat the cross-user isolation tests here as the actual proof, not a
formality.

## 3. What already exists — do not rebuild

- `backend/internal/auth/jwt.go` → `VerifyAccessToken(secret, tokenString)
  (uuid.UUID, error)`. Already written *for this ticket* (see its doc
  comment). It rejects non-HS256 algs (incl. `none` and sibling HMAC
[...]

PR #61, opened 2026-07-11
[Issue - #40] Add sqlc queries for auth and owned resources

### LINKED ISSUE #40: [Backend Migration 08] Add sqlc queries for auth and owned resources
Parent spec: #18.
Related epics: #21, #22, #23, #24, and #25.

Graphify note:

Add typed sqlc query coverage for the Go backend after migrations exist.

Acceptance criteria:
- sqlc generates typed Go accessors for users and auth token/session tables.
- sqlc generates typed Go accessors for accounts and categories CRUD/bulk behavior.
- sqlc generates typed Go accessors for transactions CRUD/list/bulk behavior.
- sqlc generates typed Go accessors or query helpers for summary aggregation.
- Queries are written to work with the authenticated DB session/RLS model.
- Generated query code is not hand edited.

Verification:
- sqlc generation command runs successfully.

GonzaloSecades:
Closes #40.

## What

The typed query layer: six hand-written SQL files in `backend/internal/db/queries/` (~35 named queries) and the sqlc-generated Go code in `backend/internal/db/gen/` (committed, never hand-edited). Plus `sql_package: "pgx/v5"` in sqlc.yaml (the pool is pgxpool), a live-gated round-trip test, README sqlc docs, and post-migration-improvements entry 0006.

No handlers, no JWT, no middleware — this is the vocabulary later tickets (#41/#43/#44+) speak.

## Query catalog

- **users.sql** — create (23505 → handler 409), get by email/id, atomic MarkUserEmailVerified, UpdateUserPassword.
- **auth_tokens.sql** — per #41's recorded decisions: create/consume for verification + reset tokens where consume is a single atomic `UPDATE ... WHERE used_at IS NULL AND expires_at > now() RETURNING user_id` (concurrent double-submit safe); invalidate-prior-on-issue; CountRecent* for rate limits; refresh create, read-only get-valid, atomic `ConsumeRefreshToken` for rotation (exactly one concurrent refresh wins), revoke, revoke-all.
- **accounts.sql / categories.sql** — CRUD with `user_id` ownership predicates, `ANY(array)` bulk delete RETURNING only deleted owned ids (fixtures: silent ignore), ORDER BY name for list determinism.
- **transactions.sql** — every read/update/delete owns rows through the accounts join (mirroring the RLS policy); list matches the fixture item shape (joined account/category names with explicit category ownership, inclusive date bounds, optional accountId via narg, date DESC); bulk create is a single `INSERT ... SELECT` from one structured `jsonb` payload via `jsonb_to_recordset` — no parallel arrays to keep in sync, JSON nulls land as SQL NULLs, and an invalid row fails the batch atomically.
- **summary.sql** — period totals (SUM CASE, overflow-safe `ABS(amount::bigint)`, COALESCE 0), category spending (owned inner join excludes uncategorized, ORDER BY value DESC), daily totals grouped by calendar day (`t.date::date`); top-3+"Other" and zero-filling stay in Go like legacy JS.

**Invariant:** ownership predicates in every owned-resource read/update/delete even though FORCE RLS is live — RLS stays the backstop, not the mechanism. The transaction INSERTs are the documented exception: handler pre-validation (legacy friendly 400/404) + RLS `WITH CHECK`, because an ownership join inside the bulk INSERT would silently drop unowned rows instead of failing loudly and atomically.

## Review history

- **[high, parity review, pre-PR]** BulkCreateTransactions couldn't represent NULL notes/category_id (sqlc emits non-nullable `[]string` for array params) — the CSV-import path would have FK-failed the whole batch. Interim `''`-as-NULL sentinel fix, superseded in round 1 by the `jsonb_to_recordset` shape above.
- **Round 1 (4 high from Gonzalo, 2 medium from Copilot), commit f601694:** atomic `ConsumeRefreshToken` rotation primitive (+ concurrent two-goroutine live test), jsonb bulk create replacing parallel unnest arrays (kills silent length-mismatch truncation), daily totals grouped by `t.date::date`, `ABS(t.amount::bigint)` overflow fix (+ boundary test at -2147483648), explicit category-join ownership predicates in ListTransactions and GetCategorySpending.
- **Round 2 (1 medium, 1 low from Copilot), commit 3586058:** README documents the insert ownership model accurately; stale sentinel reference removed from the test header.
- Divergences from the briefing where legacy won: `GetTransaction` returns own columns only (legacy GET /:id has no joined names).

## Post-migration improvements

Entry **0006** (flagged during implementation, recorded per the new rule): `transactions.amount` is 32-bit — caps a single transaction near ±2.1M ARS milliunits, a realistic figure in ARS. Ported as-is for parity; proposed bigint widening post-#27. High priority alongside 0005.

## Verification

- `sqlc generate` clean (pinned v1.31.1); `sqlc diff` proves gen/ matches queries (nothing hand-edited).
- `go vet ./...` clean; `go test ./...` green without a database (live tests skip).
- Live against PostgreSQL 17.7 (WSL): full `internal/db` suite PASS including `TestSqlcQueries_LiveDatabase` — user→account→category→transactions round trip under `set_config('app.user_id', ...)` (FORCE RLS active), ListTransactions joined shape + inclusive bounds + DESC, BulkCreateTransactions with JSON nulls + populated row + atomic failure of a batch containing an invalid row, BulkDeleteTransactions ignoring an unowned id (victim row survives), same-day different-times daily grouping, boundary amount -2147483648 through the summary queries, atomic reset-token consume (second attempt ErrNoRows), RevokeAllUserRefreshTokens — plus `TestConsumeRefreshToken_Concurrent` (two goroutines, exactly one winner). Rollback/cleanup-clean (0 leftover rows).
- graphify refreshed.


backend/internal/db/queries/auth_tokens.sql
```diff
@@ -0,0 +1,97 @@
...
+
+-- name: GetRefreshTokenByHash :one
+-- Only returns a row when the token is currently valid (not revoked, not
+-- expired); filtering happens in SQL rather than leaving expiry/revocation
+-- checks to the caller.
+SELECT *
+FROM refresh_tokens
+WHERE token_hash = sqlc.arg(token_hash)
+  AND revoked_at IS NULL
+  AND expires_at > now();
```
GonzaloSecades, line 82: [high] `GetRefreshTokenByHash` is a read-only validity check, so a refresh handler must select, then revoke, then create a replacement. Two concurrent requests can both read the token as valid before either revocation commits and can mint two valid successor sessions. Provide an atomic consume query (`UPDATE ... SET revoked_at = now() WHERE token_hash = $1 AND revoked_at IS NULL AND expires_at > now() RETURNING user_id/id`) and perform successor creation in the same transaction. Add a concurrent double-refresh test proving only one consume succeeds.

backend/internal/db/queries/transactions.sql
```diff
@@ -0,0 +1,132 @@
...
+    currencies.val
+FROM unnest(sqlc.arg(ids)::text[]) WITH ORDINALITY AS ids(val, ord)
+JOIN unnest(sqlc.arg(amounts)::integer[]) WITH ORDINALITY AS amounts(val, ord) USING (ord)
+JOIN unnest(sqlc.arg(payees)::text[]) WITH ORDINALITY AS payees(val, ord) USING (ord)
+JOIN unnest(sqlc.arg(notes)::text[]) WITH ORDINALITY AS notes(val, ord) USING (ord)
+JOIN unnest(sqlc.arg(dates)::timestamp[]) WITH ORDINALITY AS dates(val, ord) USING (ord)
+JOIN unnest(sqlc.arg(account_ids)::text[]) WITH ORDINALITY AS account_ids(val, ord) USING (ord)
+JOIN unnest(sqlc.arg(category_ids)::text[]) WITH ORDINALITY AS category_ids(val, ord) USING (ord)
+JOIN unnest(sqlc.arg(currencies)::text[]) WITH ORDINALITY AS currencies(val, ord) USING (ord)
+RETURNING *;
```
GonzaloSecades: [high] The ordinality joins silently truncate to the shortest parameter array. If any handler mapping produces arrays of different lengths, the query succeeds while dropping financial records, and `RETURNING` alone does not distinguish that from a valid shorter import. Make cardinality equality a database-enforced precondition or pass one structured JSON/composite array, and add a test showing mismatched lengths fail atomically rather than partially inserting.

backend/internal/db/queries/summary.sql
```diff
@@ -0,0 +1,50 @@
...
+SELECT
+    t.date AS date,
+    COALESCE(SUM(CASE WHEN t.amount >= 0 THEN t.amount ELSE 0 END), 0)::bigint AS income,
+    COALESCE(SUM(CASE WHEN t.amount < 0 THEN ABS(t.amount) ELSE 0 END), 0)::bigint AS expenses
+FROM transactions t
+JOIN accounts a ON a.id = t.account_id AND a.user_id = sqlc.arg(user_id)
+WHERE t.date >= sqlc.arg(start_date)
+  AND t.date <= sqlc.arg(end_date)
+  AND (sqlc.narg(account_id)::text IS NULL OR t.account_id = sqlc.narg(account_id))
+GROUP BY t.date
```
GonzaloSecades: [high] This groups by the full timestamp, not by calendar day. Two transactions on the same date at different times produce separate rows; the planned legacy-style zero-fill maps rows by `yyyy-MM-dd`, so one aggregate overwrites the other and the chart silently under-reports that day. Aggregate on an explicit day expression (`t.date::date` or the agreed timezone-aware equivalent) and cover two different timestamps on the same day.

backend/internal/db/queries/summary.sql
```diff
@@ -0,0 +1,50 @@
...
+-- computes percentage change, top-3 + "Other" category bucketing, and
+-- missing-day zero-filling in Go, matching legacy JS.
+
+-- name: GetPeriodTotals :one
+-- income/expenses/remaining are SUM(...)::bigint, coalesced to 0 so an empty
+-- period returns zeros instead of NULL (legacy JS does `value || 0`).
+SELECT
+    COALESCE(SUM(CASE WHEN t.amount >= 0 THEN t.amount ELSE 0 END), 0)::bigint AS income,
+    COALESCE(SUM(CASE WHEN t.amount < 0 THEN ABS(t.amount) ELSE 0 END), 0)::bigint AS expenses,
+    COALESCE(SUM(t.amount), 0)::bigint AS remaining
```
GonzaloSecades, line 15: [high] `ABS(t.amount)` is evaluated as `integer` before the surrounding `SUM`/cast. PostgreSQL raises `integer out of range` for the valid column value `-2147483648`, so one boundary transaction can make the financial summary fail. Cast before taking the absolute value (`ABS(t.amount::bigint)`) here and in category/daily totals, with a boundary-value live test.

GonzaloSecades (commented):
The query layer is well scoped and the generated-code workflow is clean, but I found four blocking correctness/security issues beyond Copilot's ownership comments. They affect refresh-token single use, bulk financial-record integrity, daily aggregation, and an integer boundary that can make summary queries fail. Please address the inline findings and add focused live tests for each before merging.

backend/internal/db/queries/auth_tokens.sql
GonzaloSecades: [high, fixed] f601694 — added `ConsumeRefreshToken`: single-statement `UPDATE ... SET revoked_at = now() WHERE token_hash = $1 AND revoked_at IS NULL AND expires_at > now() RETURNING id, user_id`. Exactly one of any number of concurrent refreshes wins; the winner creates the successor in the same transaction (#43/#41 wiring). `GetRefreshTokenByHash` is kept but documented as read-only-not-for-rotation. Proven by the new `TestConsumeRefreshToken_Concurrent`: two goroutines on separate pool connections race on committed rows — exactly 1 success, 1 `ErrNoRows`, PASS live.

backend/internal/db/queries/transactions.sql
GonzaloSecades: [high, fixed] f601694 — took your structured-parameter option, and it turned out cleaner than a cardinality guard: `BulkCreateTransactions` now takes a single `jsonb` array via `jsonb_to_recordset` (sqlc's analyzer accepts it, generating `payload []byte`). Length mismatch is now structurally impossible — there are no parallel arrays — and JSON nulls land as SQL NULLs natively, which also retires the previous ''-sentinel workaround. Atomicity is tested live: a batch with an id-less row fails whole (NOT NULL violation), and the valid sibling row is proven absent afterwards.

backend/internal/db/queries/summary.sql
GonzaloSecades: [high, fixed] f601694 — `GetDailyTotals` now selects and groups by `t.date::date`. You're right that grouping by raw timestamp only worked by accident of every current write path storing UTC midnight; one future non-midnight write would have silently split a day and the yyyy-MM-dd zero-fill would drop rows. Live test added: two transactions on the same day at 00:00 and 15:30 aggregate into exactly one row with summed expenses.

backend/internal/db/queries/summary.sql
GonzaloSecades: [high, fixed] f601694 — `ABS(t.amount::bigint)` in all three summary queries (period totals, category spending including its ORDER BY, daily totals). Verified the failure first: `ABS(-2147483648)` on integer raises `integer out of range` exactly as you said. Boundary live test added: a transaction with amount -2147483648 flows through GetDailyTotals and GetPeriodTotals correctly (expenses = 2147484648 with a second -1000 row).

backend/internal/db/queries/transactions.sql
```diff
@@ -0,0 +1,132 @@
...
+    c.name AS category,
+    t.category_id,
+    t.payee,
+    t.amount,
+    t.notes,
+    a.name AS account,
+    t.account_id
+FROM transactions t
+JOIN accounts a ON a.id = t.account_id AND a.user_id = sqlc.arg(user_id)
+LEFT JOIN categories c ON c.id = t.category_id
```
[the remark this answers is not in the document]
GonzaloSecades: [medium, fixed] f601694 — `ListTransactions`' category LEFT JOIN now carries `AND c.user_id = sqlc.arg(user_id)`. For well-formed data the result is identical (writes validate category ownership and RLS backstops), but the invariant is now explicit in the query: a cross-user category_id yields a NULL name instead of leaking a foreign category name.

backend/internal/db/queries/summary.sql
```diff
@@ -0,0 +1,50 @@
...
+-- Expense rows only (amount < 0), inner-joined to categories (legacy
+-- innerJoin excludes uncategorized transactions from the breakdown).
+-- Top-3 + "Other" bucketing stays in Go, as in legacy JS.
+SELECT
+    c.name AS name,
+    COALESCE(SUM(ABS(t.amount)), 0)::bigint AS value
+FROM transactions t
+JOIN accounts a ON a.id = t.account_id AND a.user_id = sqlc.arg(user_id)
+JOIN categories c ON c.id = t.category_id
+WHERE t.amount < 0
```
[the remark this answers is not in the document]
GonzaloSecades: [medium, fixed] f601694 — `GetCategorySpending`'s categories join now carries `AND c.user_id = sqlc.arg(user_id)`, same reasoning as the list query: explicit ownership on every joined owned table, so no cross-user category-name leakage even under elevated-privilege execution.

GonzaloSecades:
## Review cycle — iteration 1

**Addressed (6/6)** — Copilot 2 medium + Gonzalo 4 high, all on point, all fixed in f601694:

1. [high] Refresh rotation race → new atomic `ConsumeRefreshToken` (single UPDATE, one winner); proven by a genuinely concurrent two-goroutine live test on committed rows (1 success, 1 ErrNoRows).
2. [high] Bulk-create array truncation → replaced parallel unnest arrays with a single structured `jsonb` payload (`jsonb_to_recordset`): mismatch structurally impossible, JSON nulls native (retires the '' sentinel), invalid row fails the batch atomically (tested).
3. [high] Daily totals grouped by raw timestamp → now `t.date::date`; two same-day timestamps aggregate into one row (tested).
4. [high] `ABS(integer)` overflow at -2147483648 → `ABS(t.amount::bigint)` in all three summary queries; boundary transaction flows through summary correctly (tested).
5. [medium] `ListTransactions` category join → explicit `c.user_id` ownership predicate.
6. [medium] `GetCategorySpending` category join → same.

**Rebutted:** none — every finding survived scrutiny; two (refresh race, ABS overflow) were independently verified as real failures before fixing.

**Verification:** `sqlc generate` + `go vet` clean; full live suite PASS against PostgreSQL 17.7 including the three new tests (concurrent consume, same-day aggregation + int32-boundary summary, bulk atomic failure); DB left clean.

Copilot re-reviews on this push. 1 of 3 automated iterations used.

backend/README.md
```diff
@@ -120,24 +120,67 @@ touched by user-scoped queries.
...
+| `auth_tokens.sql` | `email_verification_tokens`, `password_reset_tokens` (atomic one-time consume, invalidate-prior, rate-limit counts), `refresh_tokens` (create, get-valid, revoke, revoke-all) |
+| `accounts.sql` | `accounts` CRUD + bulk delete |
+| `categories.sql` | `categories` CRUD + bulk delete (mirrors `accounts.sql`) |
+| `transactions.sql` | `transactions` CRUD, list (joined account/category names), bulk create, bulk delete — every query is scoped through the owning account |
+| `summary.sql` | period totals, category spending, daily totals aggregates |
+
+Every owned-resource query (accounts/categories/transactions) carries an
+explicit ownership predicate (`user_id = $N`, or a join/EXISTS through the
+owning account for transactions) even though RLS also enforces it — RLS is
+the backstop, not the mechanism (see RLS above).
```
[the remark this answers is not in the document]
GonzaloSecades: [medium, fixed] 3586058 — README adjusted to describe the actual design rather than adding predicates to the INSERTs: reads/updates/deletes carry explicit ownership; the transaction INSERTs deliberately rely on handler pre-validation (legacy friendly 400/404) + RLS WITH CHECK. Adding an ownership join inside the bulk INSERT was considered and rejected: it would *silently drop* unowned rows — a partial import — whereas WITH CHECK fails loudly and atomically, the safer failure mode for financial data. That rationale is now in the README paragraph.

backend/internal/db/queries_live_test.go
```diff
@@ -0,0 +1,560 @@
...
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgtype"
+)
+
+// TestSqlcQueries_LiveDatabase exercises a representative path through the
+// sqlc-generated query code (backend/internal/db/gen), not every query:
+// create user -> account -> category -> transactions; ListTransactions
+// (joined names, inclusive date filtering, DESC order); BulkCreateTransactions
+// (single round trip, '' NULL sentinel for notes/category_id landing as SQL
+// NULL); BulkDeleteTransactions
```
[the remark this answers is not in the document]
GonzaloSecades: [low, fixed] 3586058 — test header comment updated: it now describes the jsonb payload with JSON nulls and atomic batch failure; the '' sentinel reference was a leftover from the retired round-1 approach.

GonzaloSecades:
## Review cycle — rounds 1 and 2

**Round 1 (6 findings — 2 Copilot medium, 4 Gonzalo high), all fixed in f601694:**
1. [high] Refresh rotation race → atomic `ConsumeRefreshToken` (single UPDATE ... RETURNING), proven by a concurrent two-goroutine live test: exactly 1 winner.
2. [high] Bulk-create array cardinality truncation → single structured `jsonb` payload via `jsonb_to_recordset`; mismatch structurally impossible, JSON nulls native, invalid row fails the batch atomically (tested).
3. [high] Daily totals grouped by raw timestamp → now `t.date::date`; same-day different-times test added.
4. [high] `ABS(integer)` overflow at -2147483648 → `ABS(t.amount::bigint)` in all three summary queries; boundary live test added.
5. [medium] `ListTransactions` category join ownership predicate added.
6. [medium] `GetCategorySpending` category join ownership predicate added.

**Round 2 (2 findings), fixed in 3586058:**
7. [medium] README overclaimed ownership predicates on INSERTs → paragraph now documents the deliberate insert model (handler pre-validation + RLS WITH CHECK; an ownership join in the bulk INSERT would silently drop rows instead of failing atomically).
8. [low] Stale '' sentinel reference in the test header comment → updated for the jsonb payload.

**Verification:** `go vet` clean; full suite green; all live tests PASS against PostgreSQL 17.7 including the new concurrent-consume, boundary-amount, same-day-grouping, and batch-atomicity tests; leaked rows from a mid-development cleanup bug removed (users table back to 0).

Copilot re-reviews on this push. 2 of 3 automated iterations used.

backend/README.md
```diff
@@ -120,24 +120,74 @@ touched by user-scoped queries.
...
+`pgx.ErrNoRows`.
+
+`BulkCreateTransactions` inserts every row in one round trip from a single
+structured `jsonb` parameter (`jsonb_to_recordset`): the caller marshals the
+rows as one JSON array (fields `id`, `amount`, `payee`, `notes`, `date`,
+`account_id`, `category_id`, `currency`; dates in UTC — the `timestamp`
+cast drops any zone suffix). One parameter makes per-row integrity
+structural — there are no parallel arrays whose lengths can drift, JSON
+nulls land as SQL NULLs, and a batch containing an invalid row fails
+atomically (single INSERT statement). Refresh-token rotation must use the
```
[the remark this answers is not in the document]
GonzaloSecades: [low, fixed] PR description updated — the bulk-create bullet now describes the shipped `jsonb_to_recordset` single-payload shape (the `unnest WITH ORDINALITY` text was the superseded round-0 design), and a new Review history section traces the evolution (parity review → round 1 jsonb rewrite → round 2 docs) so reviewers can follow how the query got its final shape. Description-only; no code touched.

GonzaloSecades:
## Review cycle — round 3

**Addressed (1/1):** [low] PR description still described the superseded per-column `unnest WITH ORDINALITY` bulk-create design — updated to the shipped `jsonb_to_recordset` shape, plus a Review history section tracing the design evolution across all rounds. Description-only change; no code, no new commit.

**Rebutted:** none. This was Copilot's only round-3 finding; the code itself came back clean.

3 of 3 automated iterations used — any further Copilot rounds are Gonzalo-triggered per the cap.

GonzaloSecades:
## Registry restructure + 0006 relocation

Per Gonzalo's direction, `post-migration-improvements/` is now split into agent-owned subdirectories: `claude-backend-improvements/` (Claude's numbered parity-deviation entries — 0001–0005 moved on master, **0006 moved on this branch**) and `codex-backend-improvements/` (Codex's cross-cutting optimization program). The root README keeps the rule/template and now carries a Layout section and the full index including 0006; CLAUDE.md names the entry location and keeps each agent out of the other's directory. Master merged into this branch (only conflict: the README index — resolved by taking master's restructured version, which already lists 0006 at its new path). graphify refreshed on the merged tree.

Note: this restructure merge landed after the 3-iteration cap was reached; it contains no query/code changes (docs moves + generated artifacts only). Awaiting CI, then merge on Gonzalo's word (already given in session).

Issue #42, opened 2026-07-01
[Backend Migration 11] Implement email verification and password reset


GonzaloSecades:
Parent spec: #18.
Related epic: #22.
Spec source: `docs/specs/18-go-backend-replacement/spec.md`.

Graphify note:
- Use the current graph output to confirm impacted endpoints, flows, and business logic before implementing changes.
- Refresh the graphify artifacts before handing this ticket in or before starting another migration ticket.

Implement the email-driven auth flows for owned authentication.

Acceptance criteria:
- Registration sends an email verification token through local mail in development.
- Verification marks the user as verified after a valid one-time token is submitted.
- Password reset request sends a reset token when the email is eligible.
- Password reset accepts a valid token and new password.
- Used and expired verification/reset tokens cannot be reused.
- Local Mailpit flow is documented for manual testing.

Verification:
- `cd backend && go test ./...`
- Local Mailpit smoke covers verification and reset email delivery.

GonzaloSecades:
Merged in #65 (merge commit `1d153cd`). Branch `claude/42-email-flows` retained.

## Acceptance criteria

| Criterion | Evidence |
| --- | --- |
| Registration sends a verification token via local mail | `TestAuthLive_Register_SendsVerificationEmail_ThenVerifyEmailUnlocksLogin` + Mailpit smoke: real SMTP delivery with a working `/verify-email?token=…` link |
| Verification marks the user verified on a valid one-time token | Same test — login is 403 before and 200 after |
| Reset request sends a token when eligible | `TestAuthLive_PasswordReset_FullLifecycle`, `…_TwoConsecutiveRequests_FirstTokenInvalidated` |
| Reset accepts a valid token and new password | Full lifecycle test: old password 401, new password 200, pre-reset refresh cookie revoked |
| Used and expired tokens cannot be reused | `…_ReplayUnauthorized`, `…_ExpiredTokenUnauthorized`, `…_GarbageTokenUnauthorized`, reset replay in the lifecycle test — all 401 `INVALID_TOKEN` |
| Mailpit flow documented for manual testing | `backend/README.md` — Mailpit setup, env table, and the plain-text bodies that carry the raw token for curl testing |

## Verification

[...]

Issue #39, opened 2026-07-01
[Backend Migration 06] Create finance schema migrations with RLS


GonzaloSecades:
Parent spec: #18.
Related epic: #21.
Spec source: `docs/specs/18-go-backend-replacement/spec.md`.

Graphify note:
- Use the current graph output to confirm impacted endpoints, flows, and business logic before implementing changes.
- Refresh the graphify artifacts before handing this ticket in or before starting another migration ticket.

Port the current finance tables to SQL migrations and add the required ownership backstop with PostgreSQL RLS.

Acceptance criteria:
- Accounts, categories, and transactions preserve current table and column names where practical.
- Accounts and categories are directly owned by `user_id`.
- Transactions are owned through required `account_id`.
- Category remains optional and deletes set `category_id` to null.
- Account deletes cascade transactions.
- Transaction `amount` remains an integer milliunit value.
- Transaction `currency` is required and defaults to `ARS`.
- RLS policies restrict account/category rows by user and transaction rows through owned accounts.
- The app user setting used by Go is documented.

Verification:
- `docker compose up -d postgres`
- `cd backend && goose up`
- schema/RLS inspection or tests prove tables, defaults, foreign keys, indexes, and policies.

GonzaloSecades:
## Merged — verification evidence

Shipped in #60, merge commit `448db18`.

**What landed:** `00002_finance_base.sql` (accounts/categories/transactions ported from the Drizzle baseline: text PKs, uuid `user_id` FKs, citext per-user unique names, integer milliunit `amount`, `currency text NOT NULL DEFAULT 'ARS'`, composite `(account_id, date DESC)` index) and `00003_finance_rls.sql` (ENABLE + FORCE RLS, one fail-closed policy per table on `app.user_id`, transactions checked through the owning account).

**Acceptance criteria evidence (all live on PostgreSQL 17.7):**
- Table/column parity, FK delete rules (CASCADE / SET NULL), citext uniqueness, currency default, exact index definitions → `TestFinanceBaseSchema_LiveDatabase` PASS.
- RLS enabled+forced with exactly the expected policies → `TestFinanceRLSConfiguration_LiveDatabase` PASS (pg_class/pg_policies).
[...]

Issue #31, opened 2026-06-29
[Backend Migration 07] Add backend dev scripts and explicit DB reset workflow


GonzaloSecades:
Parent spec: #18.
Sequence: Backend Migration 07.
Depends on: #39.
Next: #40.

Graphify note:
- Use the current graph output to confirm impacted endpoints, flows, and business logic before implementing changes.
- Refresh the Graphify artifacts before handing this ticket in or before starting another migration ticket.

Add developer scripts/Make targets for the Go backend and local database workflow after the backend scaffold, local services, and initial migrations exist.

Acceptance criteria:
- Scripts or `backend/Makefile` cover backend dev, test, migration up, migration reset, and health check.
- Destructive DB reset is clearly named and never automatic.
- Root/package scripts expose convenient commands where useful without hiding destructive behavior.
- Documentation explains the normal local loop: start Postgres, run migrations, run Go API, run Next.
- This ticket does not add domain SQL queries, route handlers, auth behavior, or frontend client migration.

Verification:
- `cd backend && go test ./...`
- documented dev/test scripts run successfully
- destructive reset command is documented but not run unless explicitly requested

GonzaloSecades:
Closing as obsolete.

**Why:** this ticket's purpose in the plan was to precede #40 — `spec.md` lists NN 08 (#40, sqlc queries) as depending on NN 07 (this). That dependency is moot: #40, #41, and #42 all shipped without it. Nothing open depends on #31, so it gates nothing.

**Why it was never worked:** the `blocked` label was applied at issue creation on 2026-06-29T18:47:41Z, in the same second as `risk:low` and `type:chore` — a bulk-creation label, correct at the time because #39 had not shipped. #39 closed later and the label was never cleared. The queue protocol works the lowest open NN *not labeled blocked*, so every session since has mechanically stepped over it. That is a process gap worth noting: nothing re-evaluates a `blocked` label when its dependency closes. #30 and #27 also carry `blocked`, but both are legitimately blocked (#30 on #48, #27 on #51, both open).

**What is being given up:** backend dev scripts / Make targets, and a *named* destructive DB-reset command. Accepted — the migration has run fine without them for four tickets, and mid-migration ergonomics work is hard to justify against the remaining parity chain.

[...]

PR #68, opened 2026-07-22
[Issue - #67] Run the backend suite under the race detector in CI

### LINKED ISSUE #67: Run the backend test suite under the race detector in CI
## Problem

Nothing in this repo has ever run under the Go race detector.

#42 landed goroutine-based async email sends, `LockUser`-serialized token
issuance, and concurrent one-time-token consume paths. Those are exactly the
shapes `-race` exists to check, and none of them are exercised under it today.

#43 (Go auth middleware + DB RLS session binding) is next on the critical path
and adds middleware plus per-transaction `set_config('app.user_id', ...)`
session binding — more shared-state concurrency. That ticket should be written
*under* race coverage rather than have it retrofitted afterwards, so this lands
first.

## Why CI and not local

`-race` requires `CGO_ENABLED=1` and a C toolchain. The dev machine is Windows
with no gcc, so the race detector cannot run there. `ubuntu-latest` has gcc, so
CI is the only place this can execute.

## Scope

- Add a race-detector run of the backend suite to the `backend` job in
  `.github/workflows/ci.yml`, with `CGO_ENABLED=1`.
- Doc fix: the child-issue table in
  `docs/specs/18-go-backend-replacement/spec.md` still lists seq 07 / #31 as
  part of the active sequence. #31 was closed as obsolete (its only purpose was
  to gate #40, which shipped without it). Annotate it so the plan does not read
  as though a step were skipped.

## Acceptance criteria

- The backend CI job runs the Go test suite under `-race` against the live
  Postgres service, with `TEST_DATABASE_URL` set, so the DB-backed tests
  genuinely execute rather than skip.
- A race-detector failure fails the job.
- The spec's active-sequence table marks seq 07 / #31 as closed-obsolete.

## Verification

- CI green on the PR, with the race step's package timings showing real
  execution rather than skips.

## Risk

`risk:low` — CI config and docs. No application code changes.

If `-race` surfaces a genuine data race, that is reported rather than papered
over; fixing it is a separate decision.

GonzaloSecades:
Closes #67.

## What

Adds `go test -race ./...` to the `backend` CI job, and fixes a stale row in
the spec's active-sequence table.

## Why now

The project has never run the race detector. #42 landed goroutine-based async
email sends, `LockUser`-serialized token issuance, and concurrent
one-time-token consume paths — precisely the shapes `-race` exists to check —
and none of them are covered by it today.

#43 (auth middleware + per-transaction `set_config('app.user_id', ...)` RLS
session binding) is next on the critical path and adds more shared-state
concurrency. Landing race coverage first means that ticket gets written under
it rather than having it retrofitted.

## Second step, not a replacement

The plain `go test ./...` stays. A failure there is unambiguously a logic
failure, which keeps the signal readable; the `-race` step then adds the
concurrency coverage on top.

The usual argument for replacing rather than adding is DB-backed test time, but
the suite is small enough that it does not apply — measured locally against the
WSL Postgres:

```
ok  internal/config  1.045s
ok  internal/db      4.055s
ok  internal/http   12.602s
ok  internal/mail    1.858s
```

~20s total. Duplicating that is not worth trading the separated signal for.

## cgo

`-race` requires `CGO_ENABLED=1` and a C toolchain. The dev machine is Windows
with no gcc, so the detector cannot run there at all — CI is the only place it
executes. `CGO_ENABLED: 1` is scoped to the race step so `go vet` and the plain
test run are unaffected.

## Doc fix

The active-sequence table still listed seq 07 / #31 as a pending step. #31 was
closed as obsolete, not completed — its only role was to gate #40, and #40
shipped without it. It sat unworked because `blocked` was applied at bulk
creation and never cleared after its dependency #39 closed. The row is now
annotated, seq 08's dependency reads #39 to match what actually happened, and a
note below the table explains it so the plan does not read as though a step
were skipped.

## Verification

Pending on this PR's own CI run — that is the only place `-race` can execute.
The race step's package timings need to show real execution rather than skips;
`TEST_DATABASE_URL` is job-level `env`, so it applies to the new step.

If the detector surfaces a genuine data race, that gets reported rather than
papered over.

## Risk

`risk:low` — CI config and docs only. No application code changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,6 +74,24 @@ jobs:
         working-directory: backend
       - run: go test ./...
         working-directory: backend
+      # Kept as a second run rather than replacing the plain one above. The
+      # plain run is the readable signal — a failure there is a logic failure,
+      # unambiguously — while this run adds concurrency coverage the project
+      # has never had: the async email sends, the LockUser-serialized token
+      # issuance, and the concurrent one-time-token consume paths from #42, and
+      # from #43 onward the auth middleware and per-transaction RLS session
+      # binding. The usual argument against a second step is DB-backed test
+      # time, but the suite is ~20s total, so the duplication is not worth
+      # trading the separated signal for.
+      #
+      # The race detector needs cgo, which is off by default on the dev machine
+      # (Windows, no gcc — the detector cannot run there at all). ubuntu-latest
+      # has gcc, so this is the only place it executes.
+      - name: Test under the race detector
+        run: go test -race ./...
+        working-directory: backend
+        env:
+          CGO_ENABLED: 1
 
   openapi:
     runs-on: ubuntu-latest
--- a/docs/specs/18-go-backend-replacement/spec.md
+++ b/docs/specs/18-go-backend-replacement/spec.md
@@ -347,8 +347,8 @@ Active sequence:
 | 04 | [#37](https://github.com/GonzaloSecades/nuchi/issues/37) | Wire Go database foundation | high | #54 | no |
 | 05 | [#38](https://github.com/GonzaloSecades/nuchi/issues/38) | Create base SQL migrations with users and auth tables | high | #37 | no |
 | 06 | [#39](https://github.com/GonzaloSecades/nuchi/issues/39) | Create finance schema migrations with RLS | high | #38 | no |
-| 07 | [#31](https://github.com/GonzaloSecades/nuchi/issues/31) | Add backend dev scripts and explicit DB reset workflow | low | #39 | no while blocked |
-| 08 | [#40](https://github.com/GonzaloSecades/nuchi/issues/40) | Add sqlc queries for auth and owned resources | high | #31 | no |
+| 07 | [#31](https://github.com/GonzaloSecades/nuchi/issues/31) | Add backend dev scripts and explicit DB reset workflow | low | #39 | closed as obsolete — see note below |
+| 08 | [#40](https://github.com/GonzaloSecades/nuchi/issues/40) | Add sqlc queries for auth and owned resources | high | #39 | no |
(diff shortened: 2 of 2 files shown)


GonzaloSecades:
## Verification — CI run [29943420559](https://github.com/GonzaloSecades/nuchi/actions/runs/29943420559)

All three jobs green. The race step genuinely executed rather than skipping:

```
Test under the race detector
  env: CGO_ENABLED: 1
       TEST_DATABASE_URL: ***localhost:5432/nuchi?sslmode=disable
ok  internal/auth    3.879s
ok  internal/config  1.046s
ok  internal/db      1.679s
ok  internal/http   11.595s
ok  internal/mail    1.221s
```

Evidence it ran for real, not as a silent skip — the detector's slowdown is
visible against the plain step in the same job:

| package | `go test` | `go test -race` | factor |
| --- | --- | --- | --- |
| `internal/auth` | 1.433s | 3.879s | 2.7x |
| `internal/db` | 0.465s | 1.679s | 3.6x |
| `internal/http` | 4.738s | 11.595s | 2.4x |
| `internal/mail` | 0.207s | 1.221s | 5.9x |

A suite that skipped would report ~0.0s. Total added wall time ≈ 34s.

**No `WARNING: DATA RACE` anywhere.** #42's async email sends,
`LockUser`-serialized token issuance, and concurrent one-time-token consume
paths are clean under the detector. Nothing to escalate.

## Review status

Copilot review was requested via the API but did not attach — `reviewRequests`
came back empty, consistent with the quota exhaustion seen on #65 yesterday.
Not stalling on it. `parity-reviewer` does not apply to this diff: it reviews
migration diffs against the parity fixtures, and this PR changes no API
behavior — CI config and a docs table only.

`mergeStateStatus`: `CLEAN`.

Issue #48, opened 2026-07-01
[Backend Migration 16] Implement summary API parity


GonzaloSecades:
Parent spec: #18.
Related epic: #25.
Spec source: `docs/specs/18-go-backend-replacement/spec.md`.

Graphify note:
- Use the current graph output to confirm impacted endpoints, flows, and business logic before implementing changes.
- Refresh the graphify artifacts before handing this ticket in or before starting another migration ticket.

Implement the Go summary endpoint with current dashboard behavior preserved.

Acceptance criteria:
- Summary supports current `from`, `to`, and optional `accountId` filters.
- Date parsing and validation are strict.
- Date range limit is enforced.
- Remaining, income, expenses, and percentage changes match current behavior.
- Category breakdown groups expenses and includes `Other` behavior where applicable.
- Daily income/expense series fills missing days.
- Empty states match current frontend expectations.
- Results are scoped to the authenticated user through account ownership and RLS.
- Endpoint uses generated OpenAPI types.

Verification:
- `cd backend && go test ./...`
- Tests cover invalid dates, range limits, account filter, empty data, income/expense aggregation, category aggregation, daily series, and cross-user isolation.

Issue #44, opened 2026-07-01
[Backend Migration 12] Implement accounts API parity


GonzaloSecades:
Parent spec: #18.
Related epic: #23.
Spec source: `docs/specs/18-go-backend-replacement/spec.md`.

Graphify note:
- Use the current graph output to confirm impacted endpoints, flows, and business logic before implementing changes.
- Refresh the graphify artifacts before handing this ticket in or before starting another migration ticket.

Implement the Go accounts API with current frontend behavior preserved.

Acceptance criteria:
- List returns only the authenticated user's accounts.
- Get returns only owned accounts and 404s for missing or foreign IDs.
- Create stores an owned account and preserves duplicate user/name conflict behavior.
- Update only changes owned accounts.
- Delete only deletes owned accounts.
- Bulk delete only deletes owned account IDs.
- Endpoints use generated OpenAPI types.

Verification:
- `cd backend && go test ./...`
- Endpoint tests cover list, get, create, update, delete, bulk delete, duplicate names, unauthorized requests, and cross-user isolation.

GonzaloSecades:
## Briefing — #44 [Backend Migration 12] Implement accounts API parity

**Branch:** `claude/44-accounts-api-parity` · **PR title:** `[Issue - #44] Implement accounts API parity`

This is the first resource ticket, and the first consumer of the `RequireAuth` +
`db.WithUserTx` plumbing #43 built. #45–#48 will copy whatever shape lands here,
so the structure below is part of the deliverable, not incidental.

---

### 1. Ticket

Issue #44, acceptance criteria as written:

- List returns only the authenticated user's accounts.
- Get returns only owned accounts and 404s for missing or foreign IDs.
- Create stores an owned account and preserves duplicate user/name conflict behavior.
- Update only changes owned accounts.
- Delete only deletes owned accounts.
- Bulk delete only deletes owned account IDs.
- Endpoints use generated OpenAPI types.

Verification: `cd backend && go test ./...`, with endpoint tests covering list,
get, create, update, delete, bulk delete, duplicate names, unauthorized
requests, and cross-user isolation.

### 2. Spec section

`docs/specs/18-go-backend-replacement/spec.md`, sequence table row 12 (line 355):
seq 12, depends on #43, resource parity. #45 (categories) → #46/#47
(transactions) → #48 (summary) all depend transitively on the shape of this one.
Port-not-redesign applies: freeze observable behavior, swap the technology.

### 3. Fixtures section (the oracle)

[...]

PR #69, opened 2026-07-23
[Issue - #43] Add Go auth middleware and DB RLS session binding

### LINKED ISSUE #43: [Backend Migration 10] Add Go auth middleware and DB RLS session binding

GonzaloSecades:
Closes #43.

## What

Wires authenticated requests to the Go route layer and the PostgreSQL RLS
model. This is the ticket where **RLS stops being theoretical**: the #39
policies have existed since that ticket, but nothing set `app.user_id` at
request time, so no policy had ever engaged on a real request.

## Pieces

**`RequireAuth` middleware** (`internal/http/middleware.go`) — validates a
Bearer HS256 access token via the existing `auth.VerifyAccessToken`, puts the
verified UUID on the request context under an unexported key type, and exposes
it through `UserIDFromContext(ctx) (uuid.UUID, bool)`. The boolean makes "no
user in context" unmistakable, so a handler can't silently act as the
all-zeros user. Identity comes **only** from the `Authorization` header —
never a body field, query param, or `X-User-Id`. Applied once at a
router-level group instead of the legacy per-route `clerkMiddleware()`
repetition (a structural change that keeps observable behavior identical).

**`WithUserTx` helper** (`internal/db/rls.go`) — the single sanctioned path to
owned tables. Begins a transaction, runs `set_config('app.user_id', $1, true)`
(transaction-local, parameterized) as the first statement, runs `fn` against a
tx-bound `*dbgen.Queries`, commits on success, rolls back on error/panic.
Transaction-local is deliberate: `pgxpool` reuses connections, so a
session-level bind would leak one user's identity into the next request on that
connection. `dbgen.New(pool)` against owned tables is now forbidden — with no
transaction, the tx-local bind has nothing to attach to and the policy silently
returns zero rows. #44–#48 route through this helper.

**Expiry carve-out** (`internal/auth/jwt.go`) — `VerifyAccessToken` returns the
new `ErrAccessTokenExpired` sentinel for the expired case only (via
`errors.Is(err, jwt.ErrTokenExpired)`, checked before the generic collapse), so
the middleware can surface the contract's distinct `ACCESS_TOKEN_EXPIRED` code;
the client needs it to know when to hit `/api/auth/refresh`. Every other
failure (bad signature, wrong secret, disallowed alg, missing exp, non-UUID
subject) stays collapsed into the opaque `ErrInvalidAccessToken`.

## 401 shape — contract, not fixture

The legacy Hono handlers return flat `{"error":"Unauthorized"}`. This uses the
structured `ApiErrorResponse` envelope `{"error":{"code","message"}}` instead —
the divergence is **documented on the contract's `UnauthorizedError`
response**, so this is parity with the contract, which is the source of truth.
Only the two contract-documented codes are emitted: `UNAUTHORIZED` and
`ACCESS_TOKEN_EXPIRED`.

## Scope

Resource handlers (accounts/categories/transactions/summary) are #44–#48. The
`RequireAuth` group is mounted but intentionally empty here.

## Tests

Live DB (against Postgres 17), all green:
- Cross-user isolation proven **at the RLS policy itself** — predicate-free
  `SELECT`/`UPDATE`/`DELETE` under another user's binding, all 0 rows, so the
  policy (not a `WHERE` clause) is demonstrably doing the work.
- Fail-closed: an unbound transaction sees 0 rows, not an error and not every
  row.
- Rollback-on-error leaves no row behind.
- Missing / malformed / expired / wrong-secret / `alg:none` / HS512 → 401 with
  the right code; valid token exposes the correct UUID; body identity ignored.

Unit (no DB): the full middleware rejection matrix and the JWT expiry-vs-other
carve-out (expired satisfies `ErrAccessTokenExpired` and not
`ErrInvalidAccessToken`, and vice versa).

## Verification

Local against WSL Postgres 17 (non-superuser `nuchi` role, so the RLS proof is
real, not a superuser bypass):

```
ok  internal/auth    2.141s
ok  internal/config  0.590s
ok  internal/db      2.326s
ok  internal/http    6.630s
ok  internal/mail    1.119s
```

`go vet ./...` clean. `parity-reviewer` reviewed the diff against the contract
and fixtures — no findings; confirmed the test role is non-superuser and all
nine required scenarios are present and green. CI (with its own Postgres
service and the `-race` step from #67) is the authoritative check on this PR.

### DIFF
--- a/backend/internal/db/rls.go
+++ b/backend/internal/db/rls.go
@@ -1,0 +1,65 @@
+package db
+
+import (
+	"context"
+	"fmt"
+
+	dbgen "github.com/GonzaloSecades/nuchi/backend/internal/db/gen"
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v5/pgxpool"
+)
+
+// WithUserTx begins a transaction on pool, binds the RLS user for the
+// lifetime of that transaction, and runs fn against a *dbgen.Queries bound
+// to the transaction. It commits on success and rolls back on error or
+// panic (a panic inside fn propagates after the deferred rollback runs,
+// same as every hand-rolled tx pattern elsewhere in this codebase).
+//
+// The bind is transaction-local — set_config('app.user_id', $1, true), with
+// is_local=true as the third argument — not session-level. pgxpool reuses
+// connections across requests, so a session-level (is_local=false) bind
+// would outlive this request on its pooled connection and leak this user's
+// identity into whichever unrelated request grabs that connection next: a
+// cross-user data breach. Transaction-local scope resets automatically at
+// COMMIT or ROLLBACK, so there is nothing to leak. The user id is always
+// passed as a bind parameter (userID.String(), since set_config's value
+// argument is text), never string-interpolated into SQL.
+//
+// This is the ONLY sanctioned way to touch owned tables (accounts,
+// categories, transactions, and anything summarizing them) from here on.
+// dbgen.New(pool) called directly against the pool, outside any
+// transaction, has no transaction for a transaction-local set_config to
+// attach to — the RLS policy then sees a NULL app.user_id and silently
+// returns zero rows instead of erroring. That is not a security hole (fail
+// closed still fails closed), but it is a silent-empty-result bug that is
+// easy to ship by accident, which is why every owned-resource query must
+// route through this helper instead.
+func WithUserTx(ctx context.Context, pool *pgxpool.Pool, userID uuid.UUID, fn func(*dbgen.Queries) error) error {
+	tx, err := pool.Begin(ctx)
+	if err != nil {
+		return fmt.Errorf("db: begin user tx: %w", err)
+	}
+
+	committed := false
+	defer func() {
+		if !committed {
+			_ = tx.Rollback(ctx)
+		}
+	}()
+
+	if _, err := tx.Exec(ctx, `SELECT set_config('app.user_id', $1, true)`, userID.String()); err != nil {
+		return fmt.Errorf("db: bind rls user: %w", err)
+	}
+
+	q := dbgen.New(tx)
+	if err := fn(q); err != nil {
+		return err
+	}
+
+	if err := tx.Commit(ctx); err != nil {
+		return fmt.Errorf("db: commit user tx: %w", err)
+	}
+	committed = true
+
+	return nil
+}
--- a/backend/internal/db/rls.go
+++ b/backend/internal/db/rls.go
@@ -63,3 +63,57 @@ func WithUserTx(ctx context.Context, pool *pgxpool.Pool, userID uuid.UUID, fn fu
 
 	return nil
 }
+
+// ownedTablesRequiringRLS lists the tables whose row-level security must be
+// active for the API's connection role before the service accepts traffic.
+// Keep it in sync with the FORCE ROW LEVEL SECURITY tables in the finance
+// migrations (00003, 00004).
+var ownedTablesRequiringRLS = []string{"accounts", "categories", "transactions"}
+
+// VerifyRLSActive returns an error unless row-level security is active, for
+// the pool's connection role, on every owned table. Call it once at startup
+// and refuse to serve if it fails.
+//
+// WithUserTx binds the right app.user_id, but a binding enforces nothing when
+// the connection authenticates as a superuser or a role with BYPASSRLS: both
+// silently ignore FORCE ROW LEVEL SECURITY, so every owned-table query would
+// read or write across all tenants regardless of the binding. CI provisions an
+// ordinary role and a fresh Compose volume does too, but production credentials
+// are arbitrary and an older local volume may still have the app role as the
+// bootstrap superuser — a single such misconfiguration silently turns this
+// otherwise-correct request path into unrestricted cross-user access. This
+// guard makes that fail loud at startup instead.
+//
+// row_security_active(table) reports whether RLS would actually be enforced for
+// the current role on that table, and returns false precisely for the
+// superuser/BYPASSRLS roles this guards against — a more direct signal than
+// inspecting role attributes, which are surfaced only to make the error
+// diagnosable.
+func VerifyRLSActive(ctx context.Context, pool *pgxpool.Pool) error {
+	for _, table := range ownedTablesRequiringRLS {
+		var active bool
+		if err := pool.QueryRow(ctx, `SELECT row_security_active($1::text)`, table).Scan(&active); err != nil {
+			return fmt.Errorf("db: check row security for %q: %w", table, err)
+		}
+		if !active {
+			role, super, bypass := connectionRoleAttributes(ctx, pool)
+			return fmt.Errorf(
+				"db: row level security is not active on %q for role %q (rolsuper=%t, rolbypassrls=%t); refusing to serve because every owned-table query would bypass RLS and read or write across tenants",
+				table, role, super, bypass,
+			)
+		}
+	}
+	return nil
+}
+
+// connectionRoleAttributes fetches the current role name and its
+// superuser/bypassrls attributes for diagnostics. Best-effort: on error it
+// returns a placeholder name so the caller's error still names the failing
+// table.
+func connectionRoleAttributes(ctx context.Context, pool *pgxpool.Pool) (role string, super, bypass bool) {
+	role = "unknown"
+	_ = pool.QueryRow(ctx,
+		`SELECT current_user, rolsuper, rolbypassrls FROM pg_roles WHERE rolname = current_user`,
+	).Scan(&role, &super, &bypass)
+	return role, super, bypass
+}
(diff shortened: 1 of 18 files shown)


backend/internal/db/rls_test.go
```diff
@@ -0,0 +1,267 @@
...
+// is enforced by the RLS POLICY itself, not by an application WHERE clause.
+// WithUserTx's fn parameter only exposes *dbgen.Queries, and every
+// generated dbgen query in this codebase deliberately carries its own
+// ownership predicate, so proving the predicate-free case requires a raw
+// statement issued directly on a transaction. This test opens its own
+// transaction and issues the exact set_config('app.user_id', $1, true)
+// statement WithUserTx runs as its first statement, then probes with SQL
+// that has no user_id predicate at all — isolating the RLS policy as the
+// only thing standing between user B and user A's row.
+func TestWithUserTx_PolicyBlocksCrossUserAccess_LiveDatabase(t *testing.T) {
```
GonzaloSecades, line 127: [P1] Cover and close cross-tenant category references

The current `transactions_owner` policy only proves ownership through `account_id`. A user bound through `WithUserTx` can therefore insert or update a transaction on their own account while setting `category_id` to another user's category. PostgreSQL foreign-key checks bypass RLS, so the FK does not reject that reference (see the [PostgreSQL 17 RLS docs](https://www.postgresql.org/docs/17/ddl-rowsecurity.html)). Besides breaking tenant integrity, this creates an existence oracle and lets the other tenant's category deletion mutate this user's transaction through `ON DELETE SET NULL`. Please extend the transaction `WITH CHECK` policy to require `category_id IS NULL` or an owned category, handle any existing cross-owner rows, and add predicate-free single-create/update and bulk-create regression cases here.

backend/internal/db/rls.go
```diff
@@ -0,0 +1,65 @@
...
+// categories, transactions, and anything summarizing them) from here on.
+// dbgen.New(pool) called directly against the pool, outside any
+// transaction, has no transaction for a transaction-local set_config to
+// attach to — the RLS policy then sees a NULL app.user_id and silently
+// returns zero rows instead of erroring. That is not a security hole (fail
+// closed still fails closed), but it is a silent-empty-result bug that is
+// easy to ship by accident, which is why every owned-resource query must
+// route through this helper instead.
+func WithUserTx(ctx context.Context, pool *pgxpool.Pool, userID uuid.UUID, fn func(*dbgen.Queries) error) error {
+	tx, err := pool.Begin(ctx)
```
GonzaloSecades, line 38: [P1] Fail startup when the connection role bypasses RLS

`WithUserTx` binds the correct GUC, but it cannot enforce anything when `DATABASE_URL` authenticates as a superuser or a role with `BYPASSRLS`; both silently ignore FORCE RLS. CI and a fresh Compose volume use an ordinary role, but production credentials are arbitrary and the repo explicitly notes that older local volumes may still have `nuchi` as the bootstrap superuser. Please add a startup invariant that refuses to serve unless RLS is active for each owned table (for example via `row_security_active(...)`, optionally also checking `rolsuper`/`rolbypassrls`) and cover it with a test. Without that guard, a deployment typo turns this otherwise-correct request path into unrestricted cross-user access.

backend/internal/db/rls_test.go
GonzaloSecades: Fixed in fe01130 (migration `00004_transactions_category_owner_rls.sql`).

The `transactions_owner` `WITH CHECK` now requires `category_id IS NULL` or a category owned by the same `app.user_id`, in addition to the existing account-ownership check. The migration first nulls any pre-existing cross-owner reference — with `FORCE ROW LEVEL SECURITY` briefly lifted inside the migration's own (atomic, ACCESS EXCLUSIVE-locked) transaction, since the owner-run cleanup `UPDATE` is otherwise a silent no-op under FORCE.

This is parity, not a new rule: the legacy Hono transactions handler already rejects an unowned category with 404 "Category not found" (the `ownedCategories.length !== uniqueCategoryIds.length` check), so this is the RLS backstop for a rule the app layer already enforces — matching the "RLS is the security backstop; SQL still includes ownership predicates" invariant.

Regression coverage in `transactions_category_rls_test.go`, all predicate-free so the policy itself is proven to block: own/NULL category accepted; another tenant's category rejected on single insert, on update, and on bulk insert (whole multi-row statement atomic — the sibling good row is not inserted). Verified against a non-superuser role.

backend/internal/db/rls.go
GonzaloSecades: Fixed in fe01130.

Added `db.VerifyRLSActive(ctx, pool)`, called at startup in `cmd/api/main.go` right after the pool connects; it `os.Exit(1)`s if RLS is not active. It checks `row_security_active('<table>')` for every owned table (`accounts`, `categories`, `transactions`) — that function returns false precisely for a superuser or `BYPASSRLS` role, which is a more direct signal than reading role attributes, and the role name / `rolsuper` / `rolbypassrls` are folded into the error for diagnosis. So a `DATABASE_URL` pointing at a bypassing role fails loud at boot instead of serving every tenant's rows.

Covered both ways in `rls_test.go`: `VerifyRLSActive` passes for the ordinary application role, and is rejected for a bypassing role reached via a new `ADMIN_DATABASE_URL` (helper mirrors `liveDatabaseURL`'s skip-local / fail-in-CI contract; CI already exports `ADMIN_DATABASE_URL`, so the negative test genuinely runs there).

backend/migrations/00004_transactions_category_owner_rls.sql
```diff
@@ -0,0 +1,98 @@
...
+
+-- The cleanup UPDATE must run with FORCE RLS momentarily lifted: goose applies
+-- this migration as the ordinary table-owning role, which FORCE RLS otherwise
+-- subjects to the (unset app.user_id => zero visible rows) policy, making the
+-- UPDATE a silent no-op. The whole migration runs in one transaction and the
+-- ALTER takes an ACCESS EXCLUSIVE lock, so the lifted window is atomic and not
+-- observable to any concurrent session. A transaction's owner is its account's
+-- owner; null any category that belongs to a different user.
+-- +goose StatementBegin
+ALTER TABLE transactions NO FORCE ROW LEVEL SECURITY;
```
GonzaloSecades, line 30: [P1] Lift FORCE RLS on the cleanup's joined tables

Disabling FORCE only on `transactions` makes the target rows writable, but this `UPDATE ... FROM accounts, categories` still reads `accounts` and `categories` through their own FORCE RLS policies. Goose runs as the table-owning `nuchi` role with no `app.user_id`, so both joined tables expose zero rows and the cleanup remains a silent no-op.

I reproduced this on a disposable PostgreSQL database: migrated to v3, seeded user A's transaction on account A with user B's category (allowed by the old policy), applied v4, and `category_id` was still `category-b` rather than `NULL`. The new policy blocks future bad references, but existing ones retain the cross-tenant FK side effect and can fail later updates under the tightened `WITH CHECK`.

Please temporarily lift FORCE on `accounts` and `categories` as well (restoring all three before commit), or perform the cleanup through another controlled bypass. Add an upgrade test that seeds the invalid reference under v3, applies v4, and verifies it is nulled; the post-v4 policy tests cannot exercise this data-migration path.

GonzaloSecades: Good catch — confirmed and fixed in be1db0d.

You're exactly right: the cleanup is `UPDATE ... FROM accounts, categories`, and lifting FORCE only on `transactions` left the two joined tables subject to their own (unset `app.user_id` => zero rows) policies, so the join matched nothing and the cleanup stayed a silent no-op. I reproduced your result — seeded under v3, applied v4, `category_id` was still `migup-cat-b`.

00004 now lifts FORCE on all three tables the statement touches (transactions, accounts, categories) before the `UPDATE` and restores all three after, still inside the migration's single ACCESS EXCLUSIVE-locked transaction.

Added the upgrade test you asked for (`migration_upgrade_test.go`): it bootstraps a **throwaway database** exactly like CI (public schema owned by the app role, citext installed by the superuser, so FORCE RLS genuinely applies to the migration's role), runs the real goose migrations to v3, seeds a committed cross-owner reference the v3 policy allows, applies v4, and asserts it's nulled — the data-migration path the post-v4 policy tests can't reach. A throwaway DB (dropped WITH FORCE on teardown) keeps it from disturbing the shared `TEST_DATABASE_URL` state other packages test against in parallel; it's gated on the goose binary with the same skip-local / fail-in-CI contract as the URL helpers.

I verified it's a real regression guard, not a rubber stamp: against the transactions-only lift it fails with `got "migup-cat-b"`; against the three-table lift it passes.

Issue #45, opened 2026-07-01
[Backend Migration 13] Implement categories API parity


GonzaloSecades:
Parent spec: #18.
Related epic: #23.
Spec source: `docs/specs/18-go-backend-replacement/spec.md`.

Graphify note:
- Use the current graph output to confirm impacted endpoints, flows, and business logic before implementing changes.
- Refresh the graphify artifacts before handing this ticket in or before starting another migration ticket.

Implement the Go categories API with current frontend behavior preserved.

Acceptance criteria:
- List returns only the authenticated user's categories.
- Get returns only owned categories and 404s for missing or foreign IDs.
- Create stores an owned category and preserves duplicate user/name conflict behavior.
- Update only changes owned categories.
- Delete only deletes owned categories.
- Bulk delete only deletes owned category IDs.
- Category deletion clears `category_id` on transactions.
- Endpoints use generated OpenAPI types.

Verification:
- `cd backend && go test ./...`
- Endpoint tests cover CRUD, bulk delete, duplicates, unauthorized requests, cross-user isolation, and delete FK behavior.

GonzaloSecades:
Merged in 7a06d97 (PR #71). Graphify refreshed on master in 72e9e13.

## Verification evidence

Live tests **ran, not skipped** — confirmed with `-count=1` and `-v` so each
subtest is visible executing:

- `go test ./... -count=1` against live PostgreSQL — all packages pass,
  `internal/http` 9.4s.
- `go vet ./...` and `go build ./...` — clean.
- `bun run openapi:validate` — 28 operations, contract untouched.
- CI green on backend, openapi and frontend. The backend job re-ran the whole
  suite against its own Postgres service, so the FK and duplicate-update
  assertions are confirmed on a clean database too, and it is the only place
  `-race` runs.

Coverage, per acceptance criterion:

| Criterion | Test |
| --- | --- |
| All six operations, exact bodies | `TestCategoriesLive_HappyPath_AllOperations` |
| Empty list is `[]` not `null` | `TestCategoriesLive_ListEmpty_ReturnsEmptyArrayNotNull` |
| Duplicate on create (exact + case-differing) | `TestCategoriesLive_DuplicateName_OnCreate` |
| Duplicate on update → 409 | `TestCategoriesLive_DuplicateName_OnUpdate` |
| 404 for missing ids | `TestCategoriesLive_NotFound_Nonexistent` |
[...]

Issue #29, opened 2026-06-29
[Backend Migration 03] Define full OpenAPI contract


GonzaloSecades:
Parent spec: #18.

Graphify note:
- Use the current graph output to confirm impacted endpoints, flows, and business logic before implementing changes.
- Refresh the graphify artifacts before handing this ticket in or before starting another migration ticket.

Define the OpenAPI contract as the shared source of truth for app resources and auth endpoints. This is high-risk because it defines the API contract that Go and the frontend will generate from.

Acceptance criteria:
- OpenAPI document includes accounts, categories, transactions, summary, and auth endpoints.
- Current Hono behavior is documented before intentional changes are applied.
- Intentional changes include custom auth, Bearer access tokens, refresh cookie behavior, and transaction `currency`.
- Response wrapper remains `{ data: ... }` for app resources.
- Structured error format is documented.
- Go and TypeScript generation commands are documented.

Verification:
- OpenAPI validation command is documented and passes
- generated Go/TypeScript type command is documented or intentionally deferred

GonzaloSecades:
Merged directly to master in f5b5897 (contract) with spec sequence update in 9f6829d and graphify refresh in e4730a8.

Verification evidence:
- `bun run openapi:validate` -> `Validated openapi/nuchi.openapi.json: OpenAPI 3.1.0, 28 operation(s)`
- Contract covers health, auth (7 ops), accounts (6), categories (6), transactions (7), summary (1).
- Generation commands documented in `openapi/README.md`; execution deferred to #54 per plan.
- Graphify artifacts refreshed post-merge per ticket rule.

Next in sequence: #54 (pin generators + generate Go/TS artifacts), now unblocked.

PR #70, opened 2026-07-27
[Issue - #44] Implement accounts API parity

GonzaloSecades:
## Summary

Implements #44 (Backend Migration 12): full accounts resource parity as Go
chi handlers, establishing the `ResourceServer`/`resources.go` pattern that
#45-#48 will copy.

- `GET /api/accounts`, `POST /api/accounts`, `GET/PATCH/DELETE
  /api/accounts/{id}`, `POST /api/accounts/bulk-delete`, mounted inside the
  existing `RequireAuth` group in `router.go`.
- Every query routes through `db.WithUserTx` — no `dbgen.New(pool)` anywhere
  in the diff.
- Identity comes only from `UserIDFromContext`; a missing/unauthenticated
  context is a 401, never a zero-UUID RLS bind.
- List/get return the summary shape (`{id,name}`); create/update return the
  full row, with `plaidId` always serialized as JSON `null` (never set on
  create, per legacy behavior).
- Duplicate `(user_id, name)` maps Postgres `23505` to `409
  DUPLICATE_ACCOUNT_NAME`, with the constraint name under
  `details.constraint` (contract shape, not the legacy flat shape).
- By-id 404 covers both "missing" and "foreign" identically, via the
  single-statement UPDATE/DELETE that is both the ownership check and the
  mutation (no existence pre-check).
- Bulk-delete silently ignores missing/unowned ids and never 404s.
- Generalized `decodeAuthBody` into `decodeJSONBody[T]`; resources use a new
  64 KiB body limit (`maxResourceBodyBytes`) instead of auth's 4 KiB, to fit
  chunked bulk-delete payloads (up to 500 ids).
- Empty collections serialize as `[]`, never `null`.
- 500s use the contract's `DB_ERROR` code with the legacy per-operation
  messages ("DatabaseError - Failed to fetch accounts", etc).

## Parity note

The contract's `AccountInput.name` `minLength: 1` is stricter than legacy
Zod, which accepted `""`. This is a pre-existing, contract-sanctioned
difference from #29, not a new decision introduced here.

## Test plan

- [x] `cd backend && go vet ./...`
- [x] `cd backend && go test ./...` — full suite, including live
      `TEST_DATABASE_URL`-gated tests (WSL Postgres), all passing (see below)
- [x] `bun run openapi:validate` — contract untouched, still valid (28 ops)
- [x] `graphify update .` and refreshed `graphify-out/` committed

Live test coverage in `internal/http/accounts_live_test.go`:
happy path (all six ops, exact response bodies, summary-vs-full-row
asymmetry, `plaidId:null`), empty list `{"data":[]}`, duplicate name on
create (including citext case-insensitive) and update, 404 on
GET/PATCH/DELETE for a nonexistent id, full cross-user isolation (by-id
404s, list exclusion, mixed-owner bulk-delete), unauthorized on every route
(no header / malformed / bad signature / expired), validation (unknown
field, empty/missing name, empty ids array, empty-string id, oversized
body), and the account-delete transaction cascade.


backend/internal/http/accounts.go
```diff
@@ -0,0 +1,399 @@
...
+		return
+	}
+
+	if fields := validateAccountName(body.Name); len(fields) > 0 {
+		resp := openapi.CreateAccount400JSONResponse{ValidationErrorJSONResponse: validationError(fields...)}
+		_ = resp.VisitCreateAccountResponse(w)
+		return
+	}
+
+	id, err := uuid.NewV7()
```
GonzaloSecades: **Parity blocker — preserve the legacy resource-ID format.** The Hono handler creates account IDs with `createId()` (`app/api/[[...route]]/accounts.ts`), `ResourceId` documents legacy finance IDs as cuid-style text, and `post-migration-improvements/claude-backend-improvements/0002-finance-ids-and-uuidv7-default.md` explicitly defers UUIDv7 convergence until after parity. Returning a UUID here is an observable behavior change and creates mixed finance-ID formats without the planned data migration. Please generate a CUID2-compatible ID here and make the happy-path test assert that format; leave UUIDv7 convergence for the separate post-migration change.

backend/internal/http/resources.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/jackc/pgx/v5/pgtype"
+	"github.com/jackc/pgx/v5/pgxpool"
+)
+
+// maxResourceBodyBytes caps request bodies on the authenticated owned-
+// resource endpoints (accounts, categories, transactions, summary). Larger
+// than maxAuthBodyBytes because bulk-delete requests can legitimately carry
+// up to 500 chunked ids (lib/chunk-items.ts): a 4 KiB limit would reject
+// real client requests, not just abuse.
+const maxResourceBodyBytes = 64 * 1024
```
GonzaloSecades: **Parity blocker — do not add an undocumented 64 KiB API limit.** Neither the legacy Hono validators nor `AccountInput` / `BulkDeleteRequest` in OpenAPI declare a `maxLength`, `maxItems`, or byte-size limit. As written, a contract-valid account name or bulk-delete body just over 64 KiB now returns `400 VALIDATION_ERROR`, while Hono accepts and processes it; the frontend 500-item chunk size is a client implementation detail, not an API constraint. Please remove this resource-body cap for the parity migration (and remove the tests that freeze it), or introduce a contract/business-logic limit in a separate scoped change.

GonzaloSecades (commented):
Two migration-parity blockers are called out inline: finance resource IDs move from legacy CUID2 to UUIDv7 even though that convergence is explicitly deferred until after parity, and an undocumented 64 KiB request cap rejects payloads accepted by both the Hono implementation and the OpenAPI contract. Please preserve current behavior in this migration PR and handle either policy change separately.

backend/internal/http/accounts.go
GonzaloSecades: Fixed in b1358a1 — you're right, and this was my error in the briefing: design decision 6.6 specified UUIDv7 without reconciling it against registry entry 0002, which explicitly defers that convergence until after parity.

Adds `backend/internal/id`, a cuid2-compatible generator: 24 characters, leading lowercase letter, base36, SHA3-512 over time + random salt + process counter + host fingerprint — the same construction as @paralleldrive/cuid2, using stdlib plus the `x/crypto` already in go.mod rather than a new dependency. Unit tests cover format (100 iterations) and uniqueness (10k iterations, which is what would catch a same-millisecond collision).

`TestAccountsLive_HappyPath_AllOperations` now asserts the id matches `^[a-z][0-9a-z]{23}$` instead of merely being non-empty. A UUID fails that pattern, so a regression back to UUIDv7 fails the suite rather than passing silently.

backend/internal/http/resources.go
GonzaloSecades: Fixed in b1358a1 — agreed, and the reasoning I used for the cap (the frontend's 500-item chunk size) was exactly the client-detail-as-API-constraint mistake you named.

The cap is gone: `decodeResourceBody` decodes without `MaxBytesReader`, keeping only the strict discipline that IS contract-backed (unknown fields rejected via `additionalProperties: false`, exactly one JSON value).

Rather than just deleting the two tests that froze the cap, I inverted them, since silently dropping them would leave nothing preventing a cap from creeping back in:
- `create: large body is accepted, not capped` — posts a 128 KiB name and asserts 200 plus a full round-trip of the value.
- `bulk-delete: large id list is accepted, not capped` — posts 4000 ids and asserts 200.

Both fail if any byte cap returns. Recorded the deferred hardening as post-migration improvement 0013, which notes what the permissive behavior costs (unbounded allocation before validation on an authenticated endpoint, inherited from Hono so not a regression) and proposes doing it contract-first with real `maxLength`/`maxItems` after parity.

Issue #46, opened 2026-07-01
[Backend Migration 14] Implement transactions CRUD and list parity


GonzaloSecades:
Parent spec: #18.
Related epic: #24.
Spec source: `docs/specs/18-go-backend-replacement/spec.md`.

Graphify note:
- Use the current graph output to confirm impacted endpoints, flows, and business logic before implementing changes.
- Refresh the graphify artifacts before handing this ticket in or before starting another migration ticket.

Implement non-bulk transaction API behavior in Go.

Acceptance criteria:
- List supports `from`, `to`, and optional `accountId` filters.
- List results are scoped through owned accounts and sorted by date descending.
- Get returns only owned transactions.
- Create requires an owned account.
- Create and update allow optional category only when the category is owned.
- Update and delete affect only owned transactions.
- Amounts remain signed integer milliunits.
- `currency` is required and defaults to `ARS` where the contract specifies a default.
- Endpoints use generated OpenAPI types.

Verification:
- `cd backend && go test ./...`
- Tests cover filters, ownership, amount sign/type behavior, currency, optional category, and unauthorized requests.

GonzaloSecades:
Merged in 13d5f32 (PR #72). Graphify refreshed on master in 6fdf0d2.

## Verification evidence

Live tests **ran, not skipped** — `-count=1` throughout, and the suite was
re-run after a WSL Postgres restart so the final numbers are from a live
database:

- `go test ./... -count=1` against live PostgreSQL — all packages pass,
  `internal/http` 10.2s, `internal/ratelimit` 1.0s.
- `go vet ./...` and `go build ./...` — clean.
- `bun run openapi:validate` — 28 operations. `tsc --noEmit` and
  `bun run lint` — clean.
- CI green on backend, openapi and frontend. The backend job is the only place
  `-race` runs.

Coverage, per acceptance criterion:

| Criterion | Test |
| --- | --- |
| `from`/`to`/`accountId` filters | `TestTransactionsLive_ListFiltering`, `TestParseDateRange_*` (8 unit tests) |
| Scoped through owned accounts, date DESC | `TestTransactionsLive_ListFiltering`, `TestTransactionsLive_CrossUserIsolation` |
| Get returns only owned | `TestTransactionsLive_CrossUserIsolation` |
| Create requires an owned account | `TestTransactionsLive_ReferenceOwnership` |
| Category only when owned | `TestTransactionsLive_ReferenceOwnership` |
[...]

PR #71, opened 2026-07-29
[Issue - #45] Implement categories API parity

### LINKED ISSUE #45: [Backend Migration 13] Implement categories API parity

GonzaloSecades:
Closes #45.

Adds the six category endpoints, following the pattern #44 established. Most of this diff is deliberately symmetric with accounts; the value is in the two places categories genuinely differ.

## Endpoints

`GET|POST /api/categories`, `GET|PATCH|DELETE /api/categories/{id}`, `POST /api/categories/bulk-delete` — all inside the `RequireAuth` group, every query through `db.WithUserTx`.

## Structure

Handlers are methods on the existing `ResourceServer`, not a new type, reusing `withUser`, `withResourceID`, `decodeResourceBody`, `pgUserID`, `dbError`, `isUniqueViolation` and `constraintName` unchanged. Two helpers that were accounts-shaped but not accounts-specific moved to `resources.go` as `validateResourceName` and `logResourceError`; `accountsServerMethods` became `resourceServerMethods` and now covers both resources, so #46-#48 stay additive.

Carried over from #44 without change: identity only from `UserIDFromContext`, no `dbgen.New(pool)`, SQL ownership predicates alongside RLS, cuid2 ids via `internal/id`, no request-body cap, `[]` rather than `null` for empty collections, driver errors logged server-side only.

## The two real differences

**1. Duplicate-name PATCH returns 409, not legacy's 500.** Legacy `categories.ts` wraps its UPDATE in a bare `catch` with no `23505` branch, so renaming a category onto an existing name surfaces as a server error (fixtures line 321). `spec.md` line 104 requires this specific mismatch to be *decided in OpenAPI instead of inherited accidentally*, and `updateCategory` declares 409 with the shared `DuplicateCategoryNameError` shape. This PR ships 409 and updates post-migration improvement 0005 to record that the contract resolved it. Frozen by `TestCategoriesLive_DuplicateName_OnUpdate` (exact + case-differing).

**2. Category delete nulls transactions rather than deleting them.** `transactions.category_id` is `ON DELETE SET NULL`, the inverse of `account_id`'s `ON DELETE CASCADE`. `TestCategoriesLive_DeleteSetsTransactionCategoryNull` covers **both** delete paths and asserts the transaction *survives* with a null category — a test asserting only "the transaction is gone" would pass for the wrong reason here.

While writing that test, seeding a transaction through a bare pool was correctly rejected by RLS (`SQLSTATE 42501`, `transactions_owner`'s WITH CHECK needs a bound `app.user_id`). The seed helper now goes through `WithUserTx`, same as the request path.

## Verification

- `go build ./...`, `go vet ./...` — clean.
- `go test ./... -count=1` against live PostgreSQL — all packages pass, `internal/http` 9.4s. Every `TestCategoriesLive_*` group confirmed executing, not skipping.
- `bun run openapi:validate` — 28 operations, contract untouched (no contract change needed).

Coverage: all six operations with exact bodies, empty list as raw `{"data":[]}`, duplicates on create and update (both case variants), 404 for missing ids, cross-user isolation (by-id 404s, list exclusion, mixed-owner bulk-delete, B's row surviving), unauthorized across four failure modes on all six routes, validation including the large-body-accepted guard, and the FK behavior above.

### DIFF
--- a/backend/internal/http/accounts.go
+++ b/backend/internal/http/accounts.go
@@ -2,9 +2,7 @@ package httpapi
 
 import (
 	"errors"
-	"log/slog"
 	"net/http"
-	"unicode/utf8"
 
 	dbgen "github.com/GonzaloSecades/nuchi/backend/internal/db/gen"
 	// Aliased: the bare package name `id` would read as shadowed inside the
@@ -52,7 +50,7 @@ func (s *ResourceServer) ListAccounts(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	if err != nil {
-		logAccountsError(r, "list accounts", err)
+		logResourceError(r, "list accounts", err)
 		resp := openapi.ListAccounts500JSONResponse{DatabaseErrorJSONResponse: dbError("DatabaseError - Failed to fetch accounts")}
 		_ = resp.VisitListAccountsResponse(w)
 		return
@@ -95,7 +93,7 @@ func (s *ResourceServer) GetAccount(w http.ResponseWriter, r *http.Request, id o
 		resp := openapi.GetAccount404JSONResponse{AccountNotFoundErrorJSONResponse: accountNotFoundError()}
 		_ = resp.VisitGetAccountResponse(w)
 	default:
-		logAccountsError(r, "fetch account", err)
+		logResourceError(r, "fetch account", err)
 		resp := openapi.GetAccount500JSONResponse{DatabaseErrorJSONResponse: dbError("DatabaseError - Failed to fetch account")}
 		_ = resp.VisitGetAccountResponse(w)
 	}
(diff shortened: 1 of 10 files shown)


GonzaloSecades (commented):
Parity review: no blocking findings. CRUD and bulk auth/ownership, response shapes, CUID2 generation, conflict handling, and delete-to-NULL behavior match the Hono implementation, frozen fixtures, and OpenAPI contract. I treated PATCH duplicate-name -> 409 as the explicitly approved contract decision, not migration scope creep. Local checks passed: go vet ./..., go test -count=1 ./..., OpenAPI validation, and git diff --check.

Issue #49, opened 2026-07-01
[Backend Migration 18] Replace frontend generated client wiring


GonzaloSecades:
Parent spec: #18.
Related epic: #26.
Spec source: `docs/specs/18-go-backend-replacement/spec.md`.

Graphify note:
- Use the current graph output to confirm impacted endpoints, flows, and business logic before implementing changes.
- Refresh the graphify artifacts before handing this ticket in or before starting another migration ticket.

Replace the frontend Hono RPC client wiring with generated OpenAPI client wiring.

Acceptance criteria:
- Generated TypeScript client is wired through a small app-owned wrapper.
- Existing browser API calls remain same-origin `/api/*`.
- Bearer access tokens are attached to API calls.
- Refresh behavior is handled in the agreed client layer.
- `createApiError` behavior is preserved or intentionally mapped to the new structured error contract.
- `lib/hono.ts` is no longer used by migrated hooks.
- No feature hook public API is intentionally changed in this ticket.

Verification:
- `bun run lint`
- `bun run build`

PR #72, opened 2026-07-29
[Issue - #46] Implement transactions CRUD and list parity

### LINKED ISSUE #46: [Backend Migration 14] Implement transactions CRUD and list parity

GonzaloSecades:
Closes #46.

Two commits, reviewable independently:

1. `7402987` — widen `transactions.amount` to bigint (schema + contract + both generated sides + Drizzle + frontend guards)
2. `c369c6d` — the five non-bulk transaction endpoints

## Why the widening is here and not deferred

int4 milliunits capped a single transaction at 2,147,483.647 ARS — roughly **USD 1,400** at the BCRA reference rate, below a month's rent or a medical bill. Implementing #46 against a cap we already knew was wrong would have meant writing throwaway int32 bounds checks and tests, then deleting them.

It ships as one change because any partial version leaves two caps in one API. In particular the `jsonb_to_recordset` column list in `BulkCreateTransactions` also had to move to bigint — left at `integer`, bulk-create and the whole CSV import path would have kept the old limit while single create no longer had it, and the import path is the one users hit with a year of history.

The API bound is **narrower than the column** on purpose: ±(2^53−1), JavaScript's safe-integer limit, so every returned amount is exact in the browser.

Drizzle moved to `bigint('amount', { mode: 'number' })` — required, not cosmetic: node-postgres returns int8 as a *string*, so without it the legacy stack would start receiving `"12500"`.

Registry 0006 is closed out.

## Handlers

**Two projections, not merged.** List joins the current account/category names; get/create/update return the transaction's own columns. Renaming a category must change the next list without rewriting transactions.

**`currency` now comes from the column** — the contract requires it on `TransactionListItem` and the query didn't select it. Never hardcoded, which would lie if the data disagreed.

**Deterministic ordering.** Added `t.id DESC` as a tiebreak; most rows share midnight, so date-only ordering flapped between identical requests. Same behavior-preserving reasoning as `accounts.sql`'s `ORDER BY name` in #40.

**Dates are UTC**, computed from a single captured `now`. Legacy parses in the host process's timezone. Ported exactly otherwise, including the rule that a provided `to` does **not** re-anchor the default `from`, and the 366/367-day boundary. `from`/`to` are read raw rather than through the generated params, because `openapi_types.Date` cannot represent a malformed value — binding through it would discard precisely the inputs the three `INVALID_QUERY` messages exist for.

**Reference ownership** is validated in the same transaction as the write, account before category, matching legacy precedence. FK checks bypass RLS and prove existence, not ownership, so these lookups produce the 404s and RLS stays the backstop.

**Amounts** are bounds-checked before conversion. `minimum`/`maximum` in the contract are documentation — oapi-codegen emits no range validation — so this check is the only enforcement. Out-of-range returns 400 where legacy returned a database-driven 500; deliberate, since returning a server error for a client error we've already detected is indefensible.

**`currency` required and exactly `"ARS"`**; omitted or unknown is rejected, not defaulted. Legacy has no currency field at all, so nothing is broken.

**`categoryId`**: omitted and explicit `null` are the same outcome. This PATCH is a full replacement despite the verb, as in legacy.

**Rate limiter** is its own package — 60/60s per (user, action), mutex-guarded because Go serves requests concurrently where Node's event loop serialized the original for free, injected clock so no test sleeps, rejected attempts not recorded so a client can't lock itself out, `Retry-After` matching `max(1, ceil(ms/1000))` exactly.

## Verification

- Full backend suite green against live PostgreSQL; `go vet` clean.
- `openapi:validate` 28 operations; `bun run lint` clean; `bun test` 17/17.
- `TestMigration00005_*` seeds both int4 extremes plus an ordinary pair and zero, upgrades, proves every value survives sign-intact, then inserts 3,000,000,000 milliunits.
- Limiter tests include a 500-goroutine race that asserts exactly 60 approvals — the test that fails on an unsynchronized port.


GonzaloSecades (commented):
Superseded: re-reviewing this PR against the Go/OpenAPI target only.

backend/internal/http/transactions.go
```diff
@@ -0,0 +1,561 @@
...
+// ListTransactions implements GET /api/transactions.
+//
+// params carries accountId only. from/to are read straight from the query
+// string instead: the generated FromDate/ToDate are openapi_types.Date, a
+// struct that cannot represent a malformed value, so binding through it would
+// discard exactly the inputs the contract requires three specific
+// INVALID_QUERY messages for.
+func (s *ResourceServer) ListTransactions(w http.ResponseWriter, r *http.Request, params openapi.ListTransactionsParams) {
+	query := r.URL.Query()
+	start, end, dateErr := parseDateRange(query.Get("from"), query.Get("to"), s.now())
```
GonzaloSecades: [P2] Reject date parameters that are present but empty. `Query().Get` returns `""` for both an omitted key and `?from=`/`?to=`, so the latter currently receives the default range and returns 200. Under the OpenAPI `format: date` parameters, an explicitly empty value is malformed and should take the declared `INVALID_QUERY` 400 path; only omission should default. Preserve presence with `query.Has` (or pass presence separately) and add `from=`/`to=` cases to the date matrix.

backend/internal/http/transactions.go
```diff
@@ -0,0 +1,561 @@
...
+	if body.Currency != string(openapi.ARS) {
+		fields = append(fields, apiFieldError{Path: "currency", Message: `Currency must be "ARS".`})
+	}
+	out.currency = body.Currency
+
+	if body.Notes != nil {
+		out.notes = pgtype.Text{String: *body.Notes, Valid: true}
+	}
+	// Omitted and explicit null are the same outcome: no category.
+	if body.CategoryID != nil && *body.CategoryID != "" {
```
GonzaloSecades: [P2] Enforce the OpenAPI ID constraint for a non-null category. `categoryId` is `nullable`, but its non-null branch references `ResourceId`, which has `minLength: 1`. Treating `categoryId: ""` as no category turns an invalid client value into a successful create/update and silently clears the association. Keep `nil` as the absent/null state, but append a `categoryId` validation field error when the pointer is non-nil and empty; cover both create and update.

backend/internal/http/transactions.go
```diff
@@ -0,0 +1,561 @@
...
+	default:
+		logResourceError(r, "fetch transaction", err)
+		resp := openapi.GetTransaction500JSONResponse{DatabaseErrorJSONResponse: dbError("DatabaseError - Failed to fetch transaction")}
+		_ = resp.VisitGetTransactionResponse(w)
+	}
+}
+
+// CreateTransaction implements POST /api/transactions. Reference ownership is
+// validated inside the same transaction as the insert, so the checks and the
+// write share one RLS-bound identity and cannot interleave with a concurrent
```
GonzaloSecades: [P2] The transaction alone does not prevent a concurrent reference deletion. `validateTransactionReferences` calls plain `GetAccount`/`GetCategory` SELECTs, and under PostgreSQL READ COMMITTED those rows are not locked: a DELETE can commit after validation and before the write, making the insert/update fail with SQLSTATE 23503 and surface as a generic 500. Either use owned `SELECT ... FOR KEY SHARE` reference queries so deletion is serialized through the write, or explicitly map the FK race to the appropriate reference error; then add a coordinated concurrency regression. As written, this comment promises a guarantee the code does not provide.

GonzaloSecades (commented):
Go/OpenAPI-only review: the Goose bigint migration, `int64` amount path, RLS ownership model, CRUD SQL, and concurrent rate limiter look sound. I found three actionable P2 issues in the handwritten HTTP layer: present-but-empty date parameters bypass the declared date schema, an empty non-null `categoryId` bypasses `ResourceId.minLength`, and the reference lookups do not provide the concurrency guarantee claimed by the handler. No Hono or Drizzle behavior is used as a review criterion here.

backend/internal/http/transactions.go
GonzaloSecades: Fixed in 8d36c07 — the finding is right, but the fix lands on **404 CATEGORY_NOT_FOUND** rather than 400.

Legacy gates on `values.categoryId == null`, which catches null and undefined but **not** `""`, so an empty id falls through to its category lookup and 404s. The contract didn't decide this case explicitly (unlike the category-duplicate-update mismatch that `spec.md` line 104 calls out), so parity governs and legacy's answer is 404. Removing the `&& *body.CategoryID != ""` special case makes the Go gate identical to legacy's.

Either way your core point stands: silently clearing the association on an invalid value was wrong. Covered on both verbs by `TestTransactionsLive_EmptyCategoryIdIsARejectedReference`.

If you'd rather have 400 for the sharper client signal, say so and I'll switch it — it needs a fixtures note either way.

backend/internal/http/transactions.go
GonzaloSecades: You're right, and the comment was the actual defect — it promised a guarantee the code doesn't provide. Fixed in 8d36c07.

Took the second option you offered: `referenceErrorFromForeignKeyViolation` classifies SQLSTATE 23503 by constraint name and returns the same 404 the up-front check would have, so the race surfaces as "account not found" instead of a generic 500. The doc comment now says what the transaction actually buys (validation and write land together, or neither does) and explicitly states that it does **not** exclude a concurrent delete.

Chose the mapping over `SELECT ... FOR KEY SHARE`: taking locks on every mutation to prevent an outcome the mapping already reports correctly isn't worth the contention, and legacy has the identical unlocked race and answers it with a 500 — so 404 is strictly better than what it did rather than a regression.

On the coordinated concurrency regression: I didn't write one, deliberately. Triggering that interleaving reliably needs a lock-step coordination point inside the handler that exists only for the test, and a timing-dependent test that usually passes is worse than no test. What can actually be wrong is the classification, so `TestReferenceErrorFromForeignKeyViolation` pins it directly — including that unrelated errors (23505, 40001, a plain error, another table's FK) pass through untouched, so a genuine fault still surfaces as a 500 rather than being mislabelled a missing reference.

One fragility that deserved its own guard: the mapping switches on PostgreSQL's auto-generated constraint names, so a migration that renamed or re-created either FK would silently revert this to a 500. `TestForeignKeyConstraintNamesMatchTheSchema_LiveDatabase` asserts both names still exist.

backend/internal/http/transactions.go
GonzaloSecades: Declining this one — it would diverge from legacy, and I think the premise doesn't hold.

Legacy treats an empty value as absent: `parseStrictDate` opens with `if (!value) return null;`, so `?from=` takes the default range there exactly as it does here. There's no legacy 400 to preserve.

It would also break the current client. `features/transactions/api/use-get-transactions.ts` reads the params with `params.get('from') || ''` and passes them through unconditionally, so **every unfiltered transaction list request sends `from=&to=&accountId=`**. Rejecting empty values would 400 the default view until #49/#50 rewrote that hook — work created for no behavioral gain.

On the contract argument: `from` is an optional parameter, and OpenAPI doesn't clearly make an explicitly-empty query value malformed (`allowEmptyValue` exists for roughly this question and is deprecated precisely because its semantics are murky). Absent an explicit contract decision, parity governs, and parity says empty means absent.

Worth noting my `accountId` handling already treats empty as "no filter" for the same reason, so this keeps the three parameters consistent.

Pinned rather than left implicit: `TestTransactionsLive_EmptyDateParamsDefaultRatherThanFail` covers `?from=&to=`, `?from=`, `?to=`, and `?from=&to=&accountId=`, asserting 200 and the default window — so if anyone later decides empty *should* 400, they'll have to change a test that states the reasoning.

Happy to reverse it if you want contract strictness to win here, but it needs the frontend hook fixed in the same change.

backend/internal/http/transactions.go
```diff
@@ -0,0 +1,608 @@
...
+// ListTransactions implements GET /api/transactions.
+//
+// params carries accountId only. from/to are read straight from the query
+// string instead: the generated FromDate/ToDate are openapi_types.Date, a
+// struct that cannot represent a malformed value, so binding through it would
+// discard exactly the inputs the contract requires three specific
+// INVALID_QUERY messages for.
+func (s *ResourceServer) ListTransactions(w http.ResponseWriter, r *http.Request, params openapi.ListTransactionsParams) {
+	query := r.URL.Query()
+	start, end, dateErr := parseDateRange(query.Get("from"), query.Get("to"), s.now())
```
GonzaloSecades: [P2] This thread is still not addressed by the Go contract. The document is OpenAPI 3.0.3; for a query Parameter Object, [`allowEmptyValue` defaults to `false`](https://spec.openapis.org/oas/v3.0.3#parameter-object). `from` and `to` reference `DateString` and do not opt into empty values, so `?from=`/`?to=` cannot be normalized to omission while the contract remains unchanged. If a current caller emits empty strings, make that caller omit unset parameters (or deliberately revise the contract and regenerated artifacts); do not pin a 200 response that contradicts the published API. Preserve presence here and route explicit empties to the existing `INVALID_QUERY` 400 path.

backend/internal/http/transactions.go
```diff
@@ -0,0 +1,608 @@
...
+	if body.Notes != nil {
+		out.notes = pgtype.Text{String: *body.Notes, Valid: true}
+	}
+	// Omitted and explicit null are the same outcome: no category. Anything
+	// else — including the empty string — is treated as a reference and looked
+	// up, so it fails as a missing category rather than silently clearing the
+	// association. Legacy gates on `values.categoryId == null`, which catches
+	// null and undefined but not "", so an empty id falls through to its
+	// category lookup and 404s; matching that gate keeps the behavior.
+	if body.CategoryID != nil {
```
GonzaloSecades, line 510: [P2] This now rejects `categoryId: ""`, but with the wrong layer and response. `TransactionInput.categoryId` is nullable; its non-null branch is `ResourceId` with `minLength: 1`. The empty string is neither null nor a valid resource ID, so it must fail request validation with 400 `VALIDATION_ERROR` and a `categoryId` field error before any lookup. A 404 says a syntactically valid reference was not found and leaves the implementation out of contract. Keep only `nil` as “no category,” and update both create and update tests to assert the field-level 400.

GonzaloSecades (commented):
Follow-up review at `8d36c07`, evaluated against the Go/OpenAPI/PostgreSQL target.

The foreign-key race concern is addressed correctly: create and update both translate only SQLSTATE `23503` from the two transaction reference constraints, unrelated database errors still reach the generic error path, and the focused unit test plus live constraint-name test cover the fragile assumptions.

The other two P2 findings are not addressed against the current OpenAPI contract; I left updated inline comments. Empty date values remain disallowed by the Parameter Object, and an empty non-null `categoryId` still violates `ResourceId.minLength: 1` before reference lookup.

Verification at this head: `go test ./...`, `go vet ./...`, `go test -race ./internal/ratelimit ./internal/http`, `bun run openapi:validate`, and `git diff --check` pass. GitHub CI is green for backend, frontend, and OpenAPI.

Verdict: the concurrency finding is closed; the two input-contract findings remain actionable. This is recorded as a comment review because GitHub does not allow approving or requesting changes on one's own PR.

backend/internal/http/transactions.go
GonzaloSecades: You're right and I was wrong. Implemented in d0236c9.

The `allowEmptyValue` default settles it: it is `false` for a query Parameter Object in 3.0.3, and `from`/`to` reference `DateString` without opting in, so `?from=` is malformed and cannot be normalized to omission while the contract stands. My "murky semantics" argument confused the deprecation note with the default — the default is stated plainly. Pinning a 200 that contradicts the published contract was the wrong instinct.

Presence is now explicit: `optionalQueryParam` uses `Has`, and `parseDateRange` takes `*string` where nil means omitted and a non-nil empty value is malformed. Omission still defaults, which is the case an empty value must not be confused with.

**One extension you didn't name, flagging it for veto:** I applied the same rule to `accountId`. It references `ResourceId` (`minLength: 1`) and equally does not set `allowEmptyValue`, so `?accountId=` is malformed by the identical argument. Leaving it as "empty means no filter" would have made the three parameters behave inconsistently for no stated reason. It returns `400 INVALID_QUERY` with `accountId must not be empty.` — say the word if you'd rather scope this strictly to the dates.

Callers fixed rather than left to break at cutover: `use-get-transactions` and `use-get-summary` both passed `params.get(...) || ''` through unconditionally. They now drop unset filters via a shared `omitEmptyQueryParams` helper.

Tests inverted, not deleted — `TestParseDateRange_InvalidInputs` gains empty-from/empty-to/both-empty, and `TestTransactionsLive_EmptyQueryParamsAreRejected` replaces the test that pinned the 200 while still asserting that omission defaults.

backend/internal/http/transactions.go
GonzaloSecades: Agreed, and the layer argument is the part I got wrong. Implemented in d0236c9.

`""` is neither null nor a valid `ResourceId`, so it is a malformed request value and belongs in request validation, before any lookup. A 404 asserts that a syntactically valid reference was not found — a different claim, and one that leaves the response outside the contract. I had reached for legacy's `== null` gate as the tiebreaker, but parity doesn't get to pick the layer when the contract already classifies the value as invalid.

Now: only `nil` (omitted or explicit null) means "no category"; a present-but-empty value appends a `categoryId` field error and returns `400 VALIDATION_ERROR`.

`TestTransactionsLive_EmptyCategoryIdFailsValidation` covers create and update, asserting the status, the code, and that `details` actually names `categoryId` rather than just checking for a 400.

Worth noting the first version of that test double-read the response body — `decodeAccountsAPIError` consumes it, so asserting the status through a second helper that also decodes saw an empty body and failed with EOF. Fixed to read once. Same trap I hit earlier in this PR; it fails loudly rather than passing falsely, but it's clearly a helper shape worth being careful with.

backend/internal/http/transactions.go
```diff
@@ -0,0 +1,634 @@
...
+		return
+	}
+
+	// accountId references ResourceId (minLength 1), so an explicitly empty
+	// value is rejected for the same reason as an empty date rather than
+	// silently meaning "no filter".
+	accountID := pgtype.Text{}
+	if params.AccountId != nil {
+		if *params.AccountId == "" {
+			resp := openapi.ListTransactions400JSONResponse{InvalidDateQueryErrorJSONResponse: invalidQueryError("accountId must not be empty.")}
```
GonzaloSecades: [P3] Generalize the documented 400 response to cover this new failure. Empty `accountId` now intentionally returns `INVALID_QUERY`, but `components.responses.InvalidDateQueryError` is still described as “The date range query is invalid” and its examples cover only date errors. The runtime JSON conforms to the shared `ApiErrorResponse`, so this is not blocking, but generated API documentation now misdescribes an observable response added here. Rename/generalize that response (or at least its description), add the `accountId` example, and regenerate the contract artifacts; a clearly recorded follow-up is also reasonable if you want to keep this commit focused.

GonzaloSecades (commented):
Reviewed exact commit `d0236c903228589e3659646b580489a5e040d88e` against the Go/OpenAPI/PostgreSQL target.

Both prior P2 findings are correctly addressed:

- Query presence is now preserved with `optionalQueryParam`, `parseDateRange` distinguishes omission (`nil`) from an explicit empty value, empty `from`/`to` returns 400 `INVALID_QUERY`, and omission still selects the default range. The callers now omit unset filters instead of emitting empty values.
- A present-but-empty `categoryId` now fails request validation before reference lookup with 400 `VALIDATION_ERROR`; create and update tests assert the status, code, and `categoryId` field detail. Omitted and explicit null remain the no-category state.

Applying the same rejection to empty `accountId` is consistent with `ResourceId.minLength: 1`. I found no blocking correctness issue. I left one non-blocking P3 documentation comment because the OpenAPI 400 response is still described specifically as a date-range error even though it now also documents an `accountId` failure at runtime.

Verification: `go test ./...`, `go vet ./...`, `go test -race ./internal/ratelimit ./internal/http`, `bun test`, `bun run lint`, `bun run openapi:validate`, `bun run build`, and `git diff --check` pass. The first local build attempt reached prerendering and failed only because the workspace Clerk publishable key is a placeholder; it passed with a temporary syntactically valid test-key override. GitHub backend, frontend, and OpenAPI checks are green.

Verdict: ready from a code-correctness perspective; only the non-blocking response-documentation cleanup remains. This is a comment review because GitHub does not allow approving one's own PR.

backend/internal/http/transactions.go
GonzaloSecades: Good catch — fixed in 189dbd6 rather than deferred.

Renamed the component to `InvalidQueryError`, broadened the description to cover both malformed date ranges and empty parameters, and added `emptyParameter` and `emptyAccountId` examples alongside the date ones. Regenerated Go and TypeScript; only `transactions.go` referenced the old type name by hand, so the code churn was two lines.

Took the rename over patching just the description because #48 adds summary, which takes the same date parameters — a name scoped to dates would keep misdescribing every future non-date use of it. Better to pay it once now than to accumulate a component whose name contradicts half its usages.

Verified: openapi:validate 28 operations, full backend suite green, tsc --noEmit and lint clean.

PR #74, opened 2026-07-30
[Issue - #47] Implement transaction bulk create and delete parity

### LINKED ISSUE #47: [Backend Migration 15] Implement transaction bulk create and delete parity
Parent spec: #18.
Related epic: #24, superseded by this split ticket.
Sequence: Backend Migration 15.
Depends on: #46.
Next: #48.

Graphify note:
- Refresh the Graphify artifacts before handing this ticket in or before starting another migration ticket.

Implement transaction bulk behavior in Go, including CSV import persistence requirements.

Acceptance criteria:
- Bulk create validates every row before inserting anything.
- Bulk create validates owned account IDs and optional owned category IDs.
- No partial insert happens when any row is invalid.
- Request body and row count limits are enforced.
- Bulk delete deletes only owned transaction IDs and ignores missing or unowned IDs unless #29 intentionally changes that contract.
- Amounts remain signed integer milliunits.
- `currency` is required and defaults to `ARS` where the contract specifies a default.
- Any row-level validation error shape must be explicitly defined in #29/#36 as an intentional contract change from the current Hono behavior.

Verification:
- Tests cover invalid rows, row/body limits, cross-user IDs, rollback/no partial insert, owned-only bulk delete, and the OpenAPI-defined validation error shape.

GonzaloSecades:
Closes #47. Completes the transaction endpoints; #48 (summary) closes the backend.

## What's new beyond the obvious

**Two sqlc queries were missing.** `ListOwnedAccountIDs` and `ListOwnedCategoryIDs` — there was no owned-by-IDs lookup, only single `Get*` and `BulkDelete*`. Reference validation is set-based: distinct account and non-null category IDs, one query each, returned count vs requested. A per-row lookup at 500 rows would be 1,000 round trips inside one transaction.

**All-or-nothing is structural.** Validate rows → check references → one `INSERT ... SELECT FROM jsonb_to_recordset`, all in one `WithUserTx`. A rejected batch never reaches the statement, so there's no rollback bookkeeping to get wrong.

**Response order is guaranteed in Go, not SQL.** My first attempt used `WITH ORDINALITY` + `ORDER BY` — PostgreSQL rejects that outright (`cannot be used with a column definition list`), and even with a valid `ORDER BY`, nothing promises `RETURNING` preserves the source SELECT's order. `orderedByRequest` pairs returned rows back to the IDs this handler generated, making the correspondence a property of the code. The CSV import flow needs it to map a response row to the spreadsheet row that produced it.

**Byte limits are enforced against the stream.** `r.ContentLength` as a cheap early rejection (legacy's guard), plus `http.MaxBytesReader` against what actually arrives. The second closes legacy's hole: it trusts the header, so a chunked request bypasses the limit entirely. Tested both declared *and* undeclared oversized bodies. Recorded as improvement 0016.

`decodeBulkBody` returns a three-way outcome so `*http.MaxBytesError` maps to **413**, not the decoder's generic 400 — the one condition the contract gives a dedicated status for.

**Legacy's non-numeric `Content-Length` carve-out is unreachable in Go** and deliberately not implemented: `net/http` rejects a malformed header with its own 400 before any handler runs. A test could only pass by bypassing the server's parser, which would assert nothing.

**Transaction bulk-delete has its own ID validator.** `TransactionBulkDeleteRequest.ids` caps at 500; the accounts/categories `BulkDeleteRequest.ids` has **no** `maxItems`. Sharing one validator would silently impose a cap on two resources that deliberately lack it.

**Row errors are indexed and collected** — `[3].amount`, `[12].currency`, `$` for whole-array failures — so fixing a 500-row import doesn't take 500 round trips. Row count over the max is a 400, not a 413: `maxItems` is a schema violation, the 413 is about bytes.

## Also in here

- Removed the contract's decorative `default: "ARS"` on `CurrencyCode`, which contradicted the handler rejecting an omitted currency.
- Both bulk endpoints documented in `docs/api/transactions.md` and `docs/product/transactions.md`, including that atomicity is per *request*, not per import — a 1,200-row file is three independent batches.

## One thing I declined

I did **not** add `"currency": "ARS"` to the fixture examples. `api-parity-fixtures.md` records Hono's behavior, and legacy has no currency field anywhere — editing those examples would make the oracle describe neither stack, and it stops being an independent check on us the moment we edit it to match what we built.

## Verification

- Full backend suite green against live PostgreSQL; 27 bulk subtests. `go vet` clean.
- `openapi:validate` 28 operations; `tsc --noEmit`, `bun run lint` clean; `bun test` 17/17.
- Atomicity tested by asserting the table is **empty afterwards**, not just that the status was 4xx.


openapi/nuchi.openapi.json
```diff
@@ -1966,8 +1966,7 @@
       },
       "CurrencyCode": {
         "type": "string",
-        "description": "Transaction currency. The Go replacement starts with ARS as the required/default currency until multi-currency UX exists.",
-        "default": "ARS",
+        "description": "Transaction currency. Required on every write and currently limited to ARS: the API rejects an omitted or unknown value rather than defaulting it, so a client that believes it is sending another currency fails loudly. Multi-currency is a later change.",
```
GonzaloSecades, line 2077: [P1] Remove the remaining ARS-default claims from the contract. This schema now correctly says omission is rejected, but the same OpenAPI document still says currency defaults to ARS in `info.description` and in the list/create/bulk-create/get/update/summary operation descriptions (for example lines 6, 868, 913, and 965). Generated TypeScript documentation therefore still advertises the opposite behavior, and `docs/api/transactions.md` line 80 still says this default is only queued for removal. Please update every stale description and regenerate the Go/TypeScript outputs so the source of truth has one rule.

backend/internal/http/transactions_bulk.go
```diff
@@ -0,0 +1,359 @@
...
+
+// validateBulkTransactionInputs validates the whole batch and returns every
+// failure at once, so a client fixing a 500-row CSV import sees all the bad
+// rows rather than one per round trip.
+//
+// Field paths are indexed — `[3].amount` — with `$` for failures about the
+// array itself. The contract's ValidationError carries details.fields[].path
+// and says nothing about indices, so this is the deliberate shape choice the
+// ticket asks to be explicit about.
+func validateBulkTransactionInputs(body []transactionInputBody) ([]transactionInput, []apiFieldError) {
```
GonzaloSecades, line 327: [P1] Define the indexed batch-error format in OpenAPI, not only in implementation comments and prose docs. The shared `ValidationError` currently has a generic `email` example and leaves `details` free-form, so neither `[i].field`, `$`, nor the fact that all row errors are collected is visible to generated documentation/clients. The ticket explicitly calls this an intentional contract decision; please add a bulk-specific description/example (or schema) to the OpenAPI response and regenerate artifacts.

backend/internal/http/resources.go
```diff
@@ -37,6 +40,58 @@ func decodeResourceBody[T any](w http.ResponseWriter, r *http.Request, dst *T) b
...
+// legacy's Content-Length guard. http.MaxBytesReader then enforces it against
+// the actual stream, which is what closes the hole legacy has: a chunked
+// request carries no Content-Length, so legacy's guard passes it through and
+// streams an unbounded body into the decoder. See post-migration improvement
+// 0016 for why that counts as a deliberate divergence.
+//
+// Note legacy's other carve-out -- a non-numeric Content-Length being ignored
+// -- is unreachable here: net/http parses the header and rejects a malformed
+// value with its own 400 before any handler runs.
+func decodeBulkBody[T any](w http.ResponseWriter, r *http.Request, dst *T, limit int64) bodyDecodeOutcome {
```
GonzaloSecades, line 69: [P2] Record this stream-enforced behavior in the two OpenAPI operation descriptions. They currently describe only rejecting a numeric `Content-Length` above the limit, while this deliberate migration divergence also returns 413 when an unknown-length/chunked stream crosses the cap. Because OpenAPI is the API source of truth, generated consumers should be able to discover the behavior this helper adds.

backend/internal/http/router.go
```diff
@@ -62,10 +62,16 @@ func NewRouter(authServer *AuthServer, resources *ResourceServer) http.Handler {
 				r.Route("/api/transactions", func(r chi.Router) {
 					r.Get("/", withListTransactionsParams(resources.ListTransactions))
 					r.Post("/", resources.CreateTransaction)
+					// Static paths are registered before /{id} for readability;
+					// chi's trie prefers a static segment over a parameter
+					// regardless of declaration order, and these are POST-only
+					// while /{id} has no POST, so there is no ambiguity either
+					// way.
+					r.Post("/bulk-create", resources.BulkCreateTransactions)
```
GonzaloSecades, line 70: [P2] Extend `resourceServerMethods` with both bulk handler signatures. Its stated purpose is to grow through #48 and compile-check that handwritten handlers match the generated server interface, but it still stops at `DeleteTransaction`; these two new routes therefore bypass that guard. Adding `BulkCreateTransactions(w, r)` and `BulkDeleteTransactions(w, r)` keeps future generator changes from silently drifting from this hand-wired router.

backend/internal/http/transactions_bulk.go
```diff
@@ -0,0 +1,359 @@
...
+// ids itself, it can pair them up directly — the correspondence is then a
+// property of the code rather than an assumption, which matters for the CSV
+// import flow, where a client maps each response row back to the spreadsheet
+// row that produced it.
+//
+// Rows the database did not return are skipped rather than faked. That cannot
+// happen today (the insert is all-or-nothing), so the guard exists to fail
+// visibly as a short response instead of emitting a zero-valued transaction if
+// that ever changes.
+func orderedByRequest(requested []bulkTransactionRow, created []dbgen.Transaction) []openapi.Transaction {
```
GonzaloSecades: [P2] Please make request-order preservation part of the OpenAPI operation description as well. The implementation and API docs call this a client-visible guarantee needed by CSV import, but the authoritative bulk-create contract currently says only that all inserted rows are returned. Without the ordering statement there, a future backend/client can legally discard the behavior this helper is protecting.

GonzaloSecades (commented):
Review verdict: changes needed before merge. The ownership checks, transaction-scoped write path, bulk-delete scoping, and body-limit tests are strong. The authoritative OpenAPI contract is still internally contradictory, however, and the ticket's indexed row-error decision is not actually expressed in OpenAPI. The inline notes also identify a few guarantees and tests worth tightening while this surface is isolated.

backend/internal/http/transactions_bulk.go
```diff
@@ -0,0 +1,359 @@
...
+	byID := make(map[string]dbgen.Transaction, len(created))
+	for _, row := range created {
+		byID[row.ID] = row
+	}
+
+	ordered := make([]openapi.Transaction, 0, len(requested))
+	for _, row := range requested {
+		found, ok := byID[row.ID]
+		if !ok {
```
GonzaloSecades, line 246: [P2] Do not turn a cardinality mismatch into a committed short 200 response. `orderedByRequest` runs after `withUser` has committed, so this `continue` would both keep any partial insert and report success with fewer rows if a future trigger or query change caused `RETURNING` to omit an ID. Validate that the returned ID set exactly matches the requested set inside the transaction callback and return an error before commit; ordering can still happen afterward.

backend/internal/http/transactions_bulk_live_test.go
```diff
@@ -0,0 +1,448 @@
...
+	first.CategoryID = &categoryID
+	second := validTransactionBody(accountID)
+	second.Payee = "Salary"
+	second.Amount = 500000
+
+	rec := env.do(t, http.MethodPost, "/api/transactions/bulk-create", token, []transactionBody{first, second})
+	if rec.Code != http.StatusOK {
+		t.Fatalf("expected 200, got %d (body: %s)", rec.Code, rec.Body.String())
+	}
+	var created openapi.TransactionListResponse
```
GonzaloSecades: [P2] Decode this endpoint with `openapi.TransactionBulkCreateResponse`, not `TransactionListResponse`. These are materially different generated models: bulk-create contains `[]Transaction`, while the list response contains `[]TransactionListItem` with joined account/category fields. `encoding/json` does not enforce required fields, so decoding into the wrong model succeeds and prevents the happy-path test from pinning the endpoint's actual response contract.

backend/internal/http/transactions_bulk_live_test.go
```diff
@@ -0,0 +1,448 @@
...
+	}
+	if len(created.Data) != 2 {
+		t.Fatalf("expected 2 created rows, got %d", len(created.Data))
+	}
+
+	// Response order must match request order. The handler reorders the returned
+	// rows by the ids it generated (orderedByRequest), so this is a guarantee by
+	// construction — not an assumption that RETURNING preserves the source
+	// SELECT's order, which PostgreSQL does not promise.
+	if created.Data[0].Payee != "Market" || created.Data[1].Payee != "Salary" {
```
GonzaloSecades, line 80: [P2] Add a direct unit test that supplies `orderedByRequest` a deliberately reversed `created` slice. This live test will usually stay green even if the helper is removed because PostgreSQL currently tends to emit this simple `RETURNING` result in source order. A shuffled input to the helper is what proves the reordering guarantee rather than current executor behavior.

backend/internal/http/transactions_bulk_live_test.go
```diff
@@ -0,0 +1,448 @@
...
+		}
+	}
+
+	rec := env.do(t, http.MethodPost, "/api/transactions/bulk-create", token, bulkCreateRows(accountID, 1))
+	assertTransactionAPIError(t, rec, http.StatusTooManyRequests, "TRANSACTION_MUTATION_RATE_LIMITED")
+	if retryAfter := rec.Header().Get("Retry-After"); retryAfter != "60" {
+		t.Errorf("expected Retry-After 60, got %q", retryAfter)
+	}
+
+	// bulk-delete is a different action and untouched.
```
GonzaloSecades, line 425: [P2] Cover the other independence direction promised by the test name and ticket. This exhausts only bulk-create, then shows one bulk-delete and one single-create are still allowed; it never exhausts bulk-delete. In a fresh user/env, consume 60 bulk-delete requests, assert request 61 is 429, and verify bulk-create plus the relevant single delete/update budgets remain available. A mistakenly shared or misconfigured bulk-delete key would currently pass.

docs/product/transactions.md
```diff
@@ -143,6 +144,25 @@ arrive later without a data migration, but only ARS is accepted today. Sending
...
+- **A large import is several batches, and each is independent.** A 1,200-row
+  file is three requests, so an early batch can succeed and a later one fail.
+  The user is left with the successful batches saved. Re-importing the whole
+  file would duplicate those rows, since there is no duplicate detection yet.
+
+When a batch is rejected, every problem row is reported at once with its
+position in the file, rather than one error per attempt.
+
+There are also size caps — roughly 1 MB per import batch and 100 KB per bulk
+delete. These are far above what 500 rows of transaction data occupies; a user
```
GonzaloSecades: [P3] A size-cap hit is not necessarily malformed under the current contract. `TransactionInput.payee` and `notes` have no `maxLength`, so 500 otherwise-valid rows can legitimately exceed 1 MB (only about 2 KB of text per row). Please describe these as total-payload safeguards that long text fields can reach, rather than support guidance that labels such input malformed.

post-migration-improvements/claude-backend-improvements/0016-bulk-body-limit-enforced-against-the-stream.md
```diff
@@ -0,0 +1,70 @@
...
+implemented nor tested. A test for it could only pass by constructing a request
+that bypasses the server's parser, which would assert nothing about real
+behavior.
+
+## Proposed improvement
+
+Nothing to undo; the Go behavior is the stricter and more honest one. Two
+follow-ups worth folding into the post-parity hardening pass:
+
+- Move the byte limits into the OpenAPI document so they are discoverable
```
GonzaloSecades: [P3] This proposed follow-up is already true: the OpenAPI bulk-create and bulk-delete operation descriptions contain the exact 1,000,000/100,000-byte limits. A useful remaining improvement would be to make them machine-readable (for example, a documented vendor extension consumed by routing/middleware) or centralize them so handler and contract cannot drift; the prose descriptions should also be updated now to mention stream enforcement.

openapi/nuchi.openapi.json
GonzaloSecades: Fixed in 79647a9. You're right that this was half-done: removing the schema's `default` left six descriptions still advertising it — `info.description` plus the list/create/bulk-create/get/update/summary operations — so generated TypeScript documented the opposite of what the handler does. All rewritten to "accepted only as \ and rejected when omitted", and `docs/api/transactions.md` no longer says removal is queued; it now records that the earlier draft carried the contradiction and where it was resolved. Artifacts regenerated; there are now zero default-ARS claims in the document.

backend/internal/http/transactions_bulk.go
GonzaloSecades: Fixed in 79647a9 — this was the right call, and the ticket did explicitly ask for it.

Added a `BulkValidationError` response component used by both bulk operations. It documents `[i].field` for a row-level failure (zero-based), `$` for failures about the array itself, and that every failing row is reported in one response rather than one per request. Two examples: a multi-row failure and a whole-array one.

Previously both operations pointed at the shared `ValidationError` with its `email` example, so a generated client had no way to learn any of this. `bulkValidationError()` in the handler converts from the shared envelope, so the body shape is unchanged — only its documentation is now real.

backend/internal/http/resources.go
GonzaloSecades: Fixed in 79647a9. Both bulk operation descriptions now state that the byte limit is enforced against the request stream, not only a declared `Content-Length`, so a chunked or unknown-length body over the limit is also rejected with 413. Agreed on the principle — a divergence that only exists in a Go comment and a registry entry is invisible to every generated consumer.

backend/internal/http/router.go
GonzaloSecades: Fixed in 79647a9, and this one was a real bug rather than an omission.

The edit that was supposed to add both signatures silently failed to apply — the same scripted edit that also missed the router mount, which I caught only because the routes 404'd. So both handlers shipped outside the compile-time guard entirely. Added `BulkCreateTransactions` and `BulkDeleteTransactions` to `resourceServerMethods`.

Worth noting the guard would not have caught the original miss either, since it only fails when a listed method is absent — not when a method is missing from the list. That is a limitation of the pattern, not something this PR should change.

backend/internal/http/transactions_bulk.go
GonzaloSecades: Fixed in 79647a9. The bulk-create description now states that returned rows are in the same order as the submitted array, so a client can map each result back to the row that produced it.

Agreed on the reasoning: the guarantee is only worth having if it is binding, and a future implementation reading only the contract would otherwise be free to discard it.

openapi/nuchi.openapi.json
GonzaloSecades: Correction to the reply above — shell escaping ate a word. The rewritten wording is:

> `currency` is required and accepted only as `ARS`; an omitted or unknown value is rejected rather than defaulted.

Nothing else in that reply changed.

backend/internal/http/transactions_bulk.go
GonzaloSecades: Fixed in 79647a9 — this was the sharpest finding in the review, and it was a real bug.

`orderedByRequest` ran after `withUser` had committed, so a short `RETURNING` set would have kept the partial insert *and* answered 200 with fewer rows than requested. A silent short success on a financial write is about the worst shape that failure could take.

The cardinality check now happens inside the transaction callback, before commit: `len(out) != len(rows)` returns an error, which rolls the batch back and surfaces as a 500. The helper's skip branch is now unreachable by construction, and its comment says so — it exists to prevent a zero-valued transaction appearing in a response if that ever changes, not as a live path.

backend/internal/http/transactions_bulk_live_test.go
GonzaloSecades: Fixed in 79647a9. Decoding now uses `openapi.TransactionBulkCreateResponse`.

You're right about why this mattered: `encoding/json` ignores unknown fields and does not enforce required ones, so decoding a `[]Transaction` payload into a `[]TransactionListItem` model succeeded and silently asserted nothing. The test looked like it pinned the response contract while pinning only the fields the two models happen to share.

backend/internal/http/transactions_bulk_live_test.go
GonzaloSecades: Fixed in 79647a9 — added `TestOrderedByRequest` in `transactions_bulk_order_test.go`.

Agreed, and this is the same class of gap I flagged in #46's isolation tests: a test that passes for a reason other than the one it claims. The live assertion would have stayed green with the helper deleted, since PostgreSQL currently tends to emit this simple `RETURNING` result in source order.

The unit test feeds the helper a deliberately reversed slice, plus an already-ordered case, an empty case, and the unreachable missing-row case (asserting a skip rather than a zero-valued transaction). Deleting `orderedByRequest` now fails a test that states why.

backend/internal/http/transactions_bulk_live_test.go
GonzaloSecades: Fixed in 79647a9 — added `TestTransactionsBulkLive_BulkDeleteBudgetIsIndependent`.

You're right that the existing test's name overpromised: it exhausted only bulk-create and then checked that two other actions still worked, which does not prove bulk-delete has its own key. A bulk-delete keyed to a shared or wrong action would have passed.

The new test uses its own user, consumes 60 bulk-deletes, asserts the 61st is 429, then asserts bulk-create and single delete are both still available.

docs/product/transactions.md
GonzaloSecades: Fixed in 79647a9. You're right — `payee` and `notes` have no `maxLength`, so 500 otherwise-valid rows can legitimately cross 1 MB at roughly 2 KB of text per row.

Reworded as a total-payload cap that ordinary data stays well under but long text fields can legitimately reach, with a smaller batch as the fix rather than an implication that the data is wrong. Calling a legitimate large import "malformed" would have sent support looking for a data problem that isn't there.

post-migration-improvements/claude-backend-improvements/0016-bulk-body-limit-enforced-against-the-stream.md
GonzaloSecades: Fixed in 79647a9 — you're right, that follow-up was already done and I did not check before writing it.

Both bulk operation descriptions already carry the exact 1,000,000/100,000-byte limits, so they are discoverable by a human reading the contract. Replaced the stale item with the real remaining gap: prose cannot be consumed by routing or middleware, and the handler constants can drift from it silently. A documented vendor extension (`x-max-request-bytes`) read by the router, or a single generated constant, is what would make the two impossible to disagree.

The descriptions now also mention stream enforcement, per your P2 on `resources.go`.

openapi/nuchi.openapi.json
```diff
@@ -1349,6 +1349,54 @@
           }
         }
       },
+      "BulkValidationError": {
+        "description": "The batch is invalid. Every failing row is reported in one response rather than one per request, so a client fixing a large import does not have to discover its bad rows one at a time.\n\n`details.fields[].path` identifies the failure: `[i].field` for a problem in row `i` (zero-based), and `$` for a problem with the array itself, such as being empty or exceeding the row maximum. This indexed shape is an intentional migration decision; the current Hono validator reports the raw Zod error instead.",
```
GonzaloSecades: [P1] Do not expose this create-only error description through bulk-delete. Both operations reference `BulkValidationError`, but `validateTransactionBulkDeleteIDs` emits `ids` for empty/oversized arrays and `ids[i]` for the first empty ID—not `[i].field` or `$`, and not every failing row. Malformed JSON on either endpoint also uses this response with no `details` at all. Generated docs for bulk-delete therefore advertise a shape the handler never returns, including a transaction-create-specific whole-array example. Please use operation-specific bulk validation responses, or broaden the schema/description/examples to cover malformed bodies, create row errors, and delete ID errors explicitly; add assertions for those actual paths.

backend/internal/http/transactions_bulk.go
```diff
@@ -0,0 +1,376 @@
...
+		if err != nil {
+			// A referenced row may have been deleted since the checks above.
+			return referenceErrorFromForeignKeyViolation(err)
+		}
+		// Checked before the transaction commits, not after. A short RETURNING
+		// set means the batch did not fully land, and returning here rolls it
+		// back; discovering the mismatch after commit could only produce a
+		// successful-looking response that silently dropped rows, with the
+		// partial insert already durable.
+		if len(out) != len(rows) {
```
GonzaloSecades: [P2] Validate the returned ID set, not only its cardinality. Equal lengths do not prove every requested ID is present: a future query change, duplicate/unexpected result, or a `BEFORE INSERT` trigger that rewrites an ID can return `len(rows)` records while omitting a generated ID. `orderedByRequest` would then hit its skip branch after commit and still produce the short 200 this fix intends to prevent. Build the expected ID set inside the transaction and reject unexpected, duplicate, or missing IDs before returning nil; an equal-length wrong-ID test would pin the invariant.

docs/api/transactions.md
```diff
@@ -137,6 +143,88 @@ valid reference was missing, which is a different claim.
...
+
+Legacy also ignores a non-numeric `Content-Length`. That branch is unreachable
+here: `net/http` rejects a malformed header with its own `400` before any
+handler runs, so it is neither implemented nor tested.
+
+### Row-level validation errors
+
+Every failure in the batch is reported at once, with indexed field paths —
+`[3].amount`, `[12].currency` — and `$` for failures about the array itself
+(empty, or over the row cap). The contract's `ValidationError` carries
```
GonzaloSecades: [P3] This explanation still describes the pre-fix contract. The bulk operations no longer reference the shared `ValidationError`; they now reference `BulkValidationError`, which explicitly documents indices. Please update this paragraph to name that component and, when the contract issue above is corrected, distinguish bulk-create paths (`[i].field` / `$`) from bulk-delete paths (`ids` / `ids[i]`).

GonzaloSecades (commented):
Re-review verdict: the original 11 findings are substantively fixed, and the new tests/contract updates are mostly sound. Three follow-ups remain before I would merge: the shared bulk-validation response describes the wrong runtime shape for bulk-delete and malformed JSON, the rollback guard checks count but not ID identity, and the API prose still describes the pre-fix contract. CI is green; local backend, OpenAPI, unit, and lint checks pass. The local Next build compiled/type-checked but could not prerender because the configured Clerk publishable key is the placeholder __COPY_FROM_CURRENT_ENV_LOCAL__.

openapi/nuchi.openapi.json
GonzaloSecades: Right, and this is a defect I introduced last round — defining `BulkValidationError` fixed the *missing* contract but got its content wrong for one of the two operations referencing it.

Fixed in 82a84ae by splitting into **`BulkCreateValidationError`** and **`BulkDeleteValidationError`**, each describing only what its operation actually produces:

| Operation | Condition | path |
| --- | --- | --- |
| bulk-create | row `i` | `[i].field` |
| bulk-create | array itself | `$` |
| bulk-delete | `ids` unusable | `ids` |
| bulk-delete | first empty id | `ids[i]` |
| either | unparseable body | `details` omitted |

Chose two components over one broadened schema because the behaviors genuinely differ beyond path syntax: bulk-create **accumulates** every failing row, bulk-delete stops at the first. One description covering both would have to hedge on that, and a client can't code against a hedge. The delete description also now states explicitly that unknown and unowned ids are *not* errors — they're silently ignored — since that's the question a reader of this response would actually have.

The malformed-body case you flagged is now an example on both (`malformedBody`, no `details`), which was undocumented anywhere before.

`TestTransactionsBulkLive_ValidationErrorPathsMatchTheContract` asserts all five rows of that table against real responses.

backend/internal/http/transactions_bulk.go
GonzaloSecades: Correct, and this one stings — the check I added last round was a *proxy* for the invariant rather than the invariant. `len(out) == len(rows)` passes for an equal-sized but wrong id set, and the ordering step then silently skipped the unmatched rows and returned a committed short 200. Precisely the outcome the check was added to prevent.

Fixed in 82a84ae. `orderedByRequest` is gone; `matchCreatedToRequested` verifies and orders **in one pass**, so ordering-without-verifying is no longer expressible in the codebase. It rejects:

- an id that was requested but not returned
- an id returned that was never requested
- the same id returned twice (which would otherwise let one row satisfy two requested rows while a real row went missing)

It runs inside the transaction, so any of those roll the batch back rather than committing. There is no skip branch left to reach.

On your framing: I took "build the expected ID set" literally rather than adding a second pre-check, because the pairing map *is* the expected set — walking it to build the response and validating it are the same traversal, and splitting them is what allowed the bug in the first place.

Tests: the equal-length-wrong-id case you asked for is pinned directly (`equal length with a wrong id is an error, not a short result`), plus missing, extra, and duplicate. All unit-level, no database needed, so they run everywhere.

docs/api/transactions.md
GonzaloSecades: Fixed in 82a84ae. The paragraph named `ValidationError` and described a single indexed convention, which was stale on both counts once the components were split.

It's now a table distinguishing the two operations explicitly — bulk-create's `[i].field` / `$` from bulk-delete's `ids` / `ids[i]`, plus the omitted-`details` case for an unparseable body — and it names both new components. It also records *why* the split happened, so the next reader doesn't re-derive the shared-component mistake.

Two code comments naming the removed `orderedByRequest` went stale in the same change and are corrected too.

docs/api/transactions.md
```diff
@@ -137,6 +143,107 @@ valid reference was missing, which is a different claim.
...
+checked with one `ListOwnedAccountIDs` / `ListOwnedCategoryIDs` query each,
+comparing returned count against requested count. A per-row lookup at the
+500-row maximum would be 1,000 round trips inside one transaction. Accounts are
+checked before categories, matching the single-row precedence.
+
+### Response order is guaranteed in Go, not SQL
+
+PostgreSQL does not promise that `RETURNING` emits rows in the source SELECT's
+order, and `WITH ORDINALITY` cannot be combined with a column definition list,
+so the ordering cannot honestly be established in the query. `orderedByRequest`
```
GonzaloSecades: [P3] Keep this reference aligned with the renamed helper

The implementation replaced `orderedByRequest` with `matchCreatedToRequested`, but this paragraph still points readers at the removed symbol. Please update the name here; it would also be useful to say that the helper now verifies the exact returned ID set while restoring request order, since that invariant is the important part of the latest fix.

GonzaloSecades (commented):
Third-pass review after the latest fixes: the three prior findings are addressed, the exact-ID-set guard now runs before commit, and the operation-specific validation contracts match the handlers. I found one small documentation follow-up inline; no additional runtime or contract blockers surfaced.

docs/api/transactions.md
GonzaloSecades: Fixed in 4e0a710 — you're right, and this is the third reference to that helper. I updated two code comments when I renamed it and missed the docs one, so this paragraph was sending readers after a symbol that no longer exists.

Took the second half of your suggestion too, since it's the better framing: the section now leads with the invariant rather than only the ordering. `matchCreatedToRequested` verifies the exact returned id set *while* restoring request order, rejecting a missing id, an unrequested id, or a duplicate — and it runs inside the transaction, so any of those roll the batch back.

It also records why the two are one function, which is the part worth surviving in the docs: the earlier length-only check passed an equal-sized but wrong id set, and ordering separately then produced a committed short 200. Verifying and ordering in one pass makes ordering-without-verifying inexpressible.

Grepped the tree afterwards — no references to the old name remain outside graphify output.

Issue #51, opened 2026-07-01
[Backend Migration 20] Replace Clerk auth pages with custom auth pages


GonzaloSecades:
Parent spec: #18.
Related epics: #22 and #26.
Spec source: `docs/specs/18-go-backend-replacement/spec.md`.

Graphify note:
- Use the current graph output to confirm impacted endpoints, flows, and business logic before implementing changes.
- Refresh the graphify artifacts before handing this ticket in or before starting another migration ticket.

Replace Clerk UI routes with app-owned auth pages backed by the Go auth API.

Acceptance criteria:
- Sign-in page uses owned email/password login.
- Sign-up page uses owned registration.
- Email verification page/flow is supported.
- Request password reset page/flow is supported.
- Reset password page/flow is supported.
- Logout uses the Go logout/revoke endpoint.
- Active auth paths no longer depend on Clerk UI components.
- Auth redirects preserve the intended dashboard access behavior.

Verification:
- `bun run lint`
- `bun run build`
- Manual smoke covers sign up, verify email, sign in, refresh/session continuity, logout, and password reset.

GonzaloSecades:
Completed by #80 after integrating the finished #50 client migration.

Final evidence:

- Go-owned sign-in, sign-up, verification, password reset, refresh/session continuity, logout, and client-side dashboard guard are active
- `/api/*` now targets Go by default; every active resource hook uses the generated authenticated client
- current-head Copilot review: no new comments; 0 unresolved threads
- GitHub backend/openapi/frontend checks green
- local integrated verification: lint passed; 205 tests passed / 8 explicit live-parity skips / 0 failed; production build passed without Clerk keys; `go test ./...` passed

The retained legacy implementation is intentionally left for #27 and labeled **Future improvement — post migration** until final teardown.

Issue #50, opened 2026-07-01
[Backend Migration 19] Migrate TanStack hooks to generated client


GonzaloSecades:
Parent spec: #18.
Related epic: #26.
Spec source: `docs/specs/18-go-backend-replacement/spec.md`.

Graphify note:
- Use the current graph output to confirm impacted endpoints, flows, and business logic before implementing changes.
- Refresh the graphify artifacts before handing this ticket in or before starting another migration ticket.

Move existing server-state hooks from Hono RPC calls to the generated OpenAPI client.

Acceptance criteria:
- Accounts hooks keep their public behavior and query invalidation behavior.
- Categories hooks keep their public behavior and query invalidation behavior.
- Transactions hooks keep filters, mutation behavior, bulk behavior, and query invalidation behavior.
- Summary hook keeps its public behavior and query key shape.
- Hook internals use the generated client wrapper, not ad hoc `fetch` calls.
- No visible frontend behavior changes except intentional auth replacement work handled separately.

Verification:
- `bun run lint`
- `bun run build`
- Manual smoke covers accounts, categories, transactions, bulk transaction import path, and summary.

GonzaloSecades:
Completed by #78 and #81.

Verification evidence at the merged heads:

- accounts: 6/6 hooks on the generated authenticated client
- categories: 6/6 hooks on the generated authenticated client
- transactions: list, CRUD, bulk create/delete on the generated authenticated client
- summary: generated authenticated client
- no frontend imports of the legacy Hono client remain outside the legacy server itself
- integrated verification: `bun run lint`, 152 tests passed / 8 explicit live-parity skips / 0 failed, production build passed
- GitHub backend/openapi/frontend checks green on both merge PRs

The preserved account detail query-key mismatch is classified as **Future improvement — post migration** and is not a cutover blocker.

PR #81, opened 2026-08-09
[Issue - #50] Move transactions and summary hooks to the generated client

GonzaloSecades:
Part of #50 — the transactions and summary domains. Accounts and categories are
Codex's lane and are untouched here.

Migrates the seven transaction hooks and the summary hook from the Hono RPC
client to the generated OpenAPI client. **No hook's public API changes** — no
call site needed a signature change.

Being precise about that evidence, since an earlier version of this description
overstated it: the transport migration itself required no component edits. Later
commits on this branch *do* modify components, for the date fix described below —
`edit-transaction-sheet.tsx` and `columns.tsx` build their values through the
calendar-date helpers, and `columns.tsx` also drops a stale runtime import of the
Hono client. Those are behavioral fixes, not signature changes forced by the
migration.

## What the typed client caught

Two things the Hono path had been papering over, both now compile errors rather
than runtime surprises:

- **`currency` is required** and accepted only as `ARS`. The legacy Hono route
  has no currency field at all, so nothing upstream supplies one.
- **`date` on write is a calendar date** (`yyyy-MM-dd`), not an instant. The
  form holds a `Date`.

Both are handled in `transaction-payload.ts`, at the transport boundary, rather
than by changing the form. The date is formatted from **local** parts via
date-fns: `toISOString().slice(0, 10)` would convert to UTC first and file a
transaction picked on the 9th as the 8th for any user east of Greenwich. The
CSV import path already emits `yyyy-MM-dd`, so it passes through untouched and
only gains the currency.

## Bulk errors now name the row

Bulk failures arrive with indexed paths — `[3].amount` for a row, `$` for the
whole array, `ids`/`ids[i]` for bulk-delete — and a body that could not be
parsed carries no `details` at all. `bulk-errors.ts` decodes every one of those
forms, and the import screen reports the failing row instead of a flat "failed
to import".

Row numbers are offset by however many rows already succeeded, because the
import submits in chunks of 500: an error at index 3 of the second chunk is row
**504** of the user's file, and reporting it as row 4 would send them to the
wrong line.

## Two deliberate behavior changes, both in the import path

- `useBulkCreateTransactions` no longer raises its own toast. The import screen
  is its only caller and already catches the failure to show the indexed
  message; the hook's generic toast fired alongside it, so users got two toasts
  and the useful one was second.
- `useBulkDeleteTransactions` surfaces array-level messages (notably the 500-id
  cap, which transactions enforce and accounts/categories deliberately do not)
  instead of a fixed string. Per-row indexes are filtered out there — that flow
  is selection-driven, so the user has no row numbers to map them to.

Flagging both since "no public API changes" was the bar; these are observable,
contained to one caller each, and leave the user better informed.

## Verification

`bun run lint` clean, `bun test` 94 pass / 0 fail (26 new), `bunx tsc --noEmit`
matches the master baseline, `bun run build` succeeds.

Live-checked against the running Go API: `GET /api/transactions` returns 200
through the migrated hook and the list renders with amounts correctly converted
from milliunits. The exact payload shapes both adapters produce — single create
and bulk create — were posted to the real API and accepted, which is what
validates the currency, date and null-default decisions.

Because nothing on master populates the access-token store yet, end-to-end UI
verification was done on a throwaway local merge with #51; that merge is not
pushed and neither branch depends on the other.

## Found while verifying: a date off-by-one between the stacks

Not caused by this PR — it is a backend serialization difference — but it is
user-visible and worth its own decision. Same row, same database:

| | value |
| --- | --- |
| stored | `2026-08-07 00:00:00` (`timestamp without time zone`) |
| Hono returns | `2026-08-07T03:00:00.000Z` |
| Go returns | `2026-08-07T00:00:00Z` |
| table shows (UTC-3) | Hono **07 August**, Go **06 August** |

node-postgres reads a naive timestamp as local; Go reads it as UTC. Every
transaction date renders a day early for any user west of Greenwich, which
includes Argentina. Registry entry 0014 covers the *filter* side of this and
"Known gaps" notes that `date` models a calendar date as a timestamp, but the
response side does not appear to be recorded. Raising rather than touching it —
`backend/**` is frozen and this is your call.

### DIFF
--- a/app/(dashboard)/transactions/columns.tsx
+++ b/app/(dashboard)/transactions/columns.tsx
@@ -11,6 +11,8 @@ import { Checkbox } from '@/components/ui/checkbox';
 import { client } from '@/lib/hono';
 import { formatCurrency } from '@/lib/utils';
 import { format } from 'date-fns';
+
+import { localDateFromCalendarDate } from '@/features/transactions/api/transaction-date';
 import { AccountColumn } from './account-column';
 import { Actions } from './actions';
 import { CategoryColumn } from './category-column';
@@ -57,8 +59,14 @@ export const columns: ColumnDef<ResponseType>[] = [
       );
     },
     cell: ({ row }) => {
-      const date = row.getValue('date') as Date;
-      return <span>{format(date, 'dd MMMM, yyyy')}</span>;
+      // The hook normalizes this to `yyyy-MM-dd`. Rendering it through the
+      // calendar-date helper keeps the displayed day the stored one; handing
+      // the raw value to `format` re-reads it as an instant and shows the
+      // previous day for any viewer west of Greenwich.
+      const date = row.getValue('date') as string;
+      return (
+        <span>{format(localDateFromCalendarDate(date), 'dd MMMM, yyyy')}</span>
+      );
     },
   },
   {
--- a/app/(dashboard)/transactions/columns.tsx
+++ b/app/(dashboard)/transactions/columns.tsx
@@ -1,26 +1,44 @@
 'use client';
 
-import { InferResponseType } from 'hono';
-
 import { ColumnDef } from '@tanstack/react-table';
 import { ArrowUpDown } from 'lucide-react';
 
 import { Badge } from '@/components/ui/badge';
 import { Button } from '@/components/ui/button';
 import { Checkbox } from '@/components/ui/checkbox';
-import { client } from '@/lib/hono';
+import type { components } from '@/lib/api/generated/schema';
 import { formatCurrency } from '@/lib/utils';
 import { format } from 'date-fns';
 
-import { localDateFromCalendarDate } from '@/features/transactions/api/transaction-date';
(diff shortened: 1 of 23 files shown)


GonzaloSecades:
### Codex review iteration 1 — ready to merge

The final medium finding is addressed: `useBulkCreateTransactions` no longer emits one success toast per 500-row chunk; the import page remains the single owner of completion/error messaging while transaction and summary invalidations still run after each mutation.

The branch was then integrated with current `master` (including #78, #79, and #82) and verified as a combined tree:

- `bun run lint` — passed
- `bun test` — 152 passed, 8 explicit live-parity skips, 0 failed
- `bun run build` — passed with CI environment values
- GitHub `backend`, `openapi`, and `frontend` checks — green
- latest Copilot code review — no new findings
- all review threads — resolved

Gonzalo approved the merge orchestration in this session.

PR #82, opened 2026-08-09
Add differential parity harness comparing Hono and Go

GonzaloSecades:
Ticket C of this batch. No issue number — following the precedent of
`claude/investigation-argentina-openbanking` for non-ticket work. Happy to file
one if you'd rather it be tracked.

Touches nothing existing: one new directory, `tests/parity/`.

## Why now

**This can only be built while both stacks exist.** After #27 removes Hono
there is nothing left to compare against, and any divergence still present just
becomes how the app behaves.

Two oracles already existed and neither compares the stacks. The fixtures
record Hono, but nothing executes them. The Go suite asserts Go against the
contract. Both pass while disagreeing with each other — so a difference the
contract does not pin down (a timezone, an ordering tiebreak, a rounding step)
is invisible to both until `USE_GO_API` flips and it becomes a cutover-day bug.

## It found one immediately

`transactions.date` is `timestamp without time zone`. Both stacks read the same
bytes and disagree about what timezone those bytes are in:

| | value |
| --- | --- |
| stored | `2026-08-07 00:00:00` |
| node-postgres (Hono) | `2026-08-07T03:00:00.000Z` |
| Go | `2026-08-07T00:00:00Z` |

A browser renders those as **different calendar days** anywhere west of
Greenwich. Every transaction date shows a day early — including for Argentina,
the app's own market. I hit this in the real UI while verifying #50 before I
understood the cause.

Entry 0014 covers the *filter* side of the same modelling problem, and
`docs/api/transactions.md` notes under "Known gaps" that `date` is a `timestamp`
modelling what is really a calendar date. The response side does not appear to
be recorded anywhere. I have not touched it — `backend/**` is frozen and it is
your call.

**The test asserts the difference is present rather than asserting the two
agree.** An assertion that fails today, inside a freeze that forbids fixing it,
gets ignored and then deleted. Written this way it documents the defect, proves
it is still there, and fails loudly the day someone changes either side.

## Known divergences as data

`divergences.ts` lists the deliberate ones, each pointing at where it is argued
(a numbered registry entry, or a `docs/api/` "Divergences" section). Without
that list a run is a wall of red someone has to re-triage by hand every time,
and the one real regression hides among decisions we already made on purpose.
Anything reported that is **not** on the list is a finding. The date divergence
is deliberately marked `expected: false`.

## Scope, and its limit

| Stack | How it is driven |
| --- | --- |
| Go | Over HTTP, as a real client, with a real session |
| Hono | Through node-postgres, the driver its Drizzle setup uses |

The Go side is fully end-to-end: register, verify by SQL (the emailed token is
stored hashed, so depending on a mail catcher would tie a data comparison to
SMTP), log in, call with the Bearer token.

**The Hono side is the honest limitation.** Its routes authenticate through
Clerk, which is not scriptable from a test, so this compares at the data layer
rather than through Hono's handlers. That is faithful for how stored values
become JSON — the handler passes them through — but would *not* catch a
difference introduced inside a Hono handler, such as an ordering or aggregation
choice. Closing that means mocking `@hono/clerk-auth` and calling the app
in-process, or a Clerk test session; both viable, neither needed for what this
found first. The seam is documented and the support layer is separate for
exactly that reason.

So this delivers the date/serialization comparison properly and leaves list
ordering and summary aggregation as documented next steps rather than
half-built.

## Verification

`bun run lint` clean, `bunx tsc --noEmit` matches the master baseline,
`bun run build` compiles. `bun test` at the root is 70 pass / 6 skip / 0 fail —
the parity tests skip in the default run and do not affect CI.

Run against a live Postgres and a live Go API, all 3 assertions pass. Every test
**skips rather than passes** when it cannot reach both stacks: a differential
test that silently passes because it reached nothing is worse than no test,
because it looks like evidence.

### DIFF
--- a/tests/parity/README.md
+++ b/tests/parity/README.md
@@ -1,0 +1,120 @@
+# Differential parity harness
+
+Drives the same input at both stacks and compares what comes back.
+
+**This can only be built while both stacks exist.** After
+[#27](https://github.com/GonzaloSecades/nuchi/issues/27) removes Hono there is
+nothing left to compare against, and any divergence still present becomes
+simply "how the app behaves". That is the whole reason this lands before the
+cutover rather than after it.
+
+## The gap it closes
+
+Two oracles already exist and neither one compares the stacks:
+
+- `docs/specs/18-go-backend-replacement/api-parity-fixtures.md` records what
+  Hono does. It is a document; nothing executes it.
+- The Go test suite asserts Go against the contract.
+
+Both pass while disagreeing with each other. A difference that the contract does
+not pin down — a timezone, an ordering tiebreak, a rounding step — sits in the
+gap between them, invisible to either, until `USE_GO_API` flips and it becomes a
+production bug on cutover day.
+
+That is not hypothetical: the first thing this harness was pointed at, it found
+one. See "Date serialization" below.
+
+## Running it
+
+Needs Postgres and a Go API. From the repo root:
+
+```bash
+wsl -d Ubuntu-20.04 -u root -e sh -c "service postgresql start; sleep 7200"
+```
+
+Then the API, from `backend/`:
+
+```bash
+DATABASE_URL='postgres://nuchi:nuchi@localhost:5432/nuchi?sslmode=disable' AUTH_JWT_SECRET='local-parity-secret-at-least-32-bytes-long' APP_BASE_URL='http://localhost:3000' SMTP_ADDR='localhost:1025' MAIL_FROM='nuchi@localhost' go run ./cmd/api
+```
+
+Then the harness:
+
+```bash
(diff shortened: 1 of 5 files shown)


GonzaloSecades:
### Codex merge review — ready

The final readiness-probe finding is addressed at `3bd57b7`: the harness now requires both the expected `401` and `INVALID_REFRESH_TOKEN` envelope, so an unrelated service cannot masquerade as the Go API.

- all 6 review threads resolved
- latest Copilot review covers the current head with no new findings
- `backend`, `openapi`, and `frontend` checks green
- intentionally deferred as **Future improvement — post migration**: live CI execution, in-process Hono route coverage, and additional ordering/summary comparisons

Gonzalo explicitly approved orchestrating and merging this batch.

Issue #90, opened 2026-08-09
Retire the differential parity harness and drop the pg dependency


GonzaloSecades:
**Decided:** the harness is retired. Target end state is **Next + Go, nothing else**, and this is the last non-Go database driver in the frontend package.

**Depends on #87** — the divergence must be recorded in the registry before this deletes the only executable record of it.

## Why

`tests/parity/` compares the legacy stack against Go. Its own README states the premise: it can only exist while both stacks do, and after teardown "any divergence still present becomes simply how the app behaves".

Once #84 and #85 land, the legacy stack is gone. The harness's comparator is `node-postgres` reading the database directly — that stood in for Hono's data layer, and with Hono deleted it represents nothing the app uses. The test would still pass; it would just no longer be a parity test.

`pg` is also the only remaining non-Go database driver in `package.json`.

## Scope

- Delete `tests/parity/**` (README, `divergences.ts`, `divergences.test.ts`, `date-serialization.test.ts`, `support/stacks.ts`).
- Remove `pg` from `package.json`. Confirm nothing else imports it — `db/connection.ts` did, but that is deleted in #85, which lands first and deliberately leaves `pg` alone for you.

## Hard requirement

**Do not start until #87 has merged.** The harness currently holds the only executable record of the date-serialization divergence — Go serializes a naive `timestamp` as UTC, node-postgres as host-local, and a browser west of Greenwich renders those as different calendar days. #87 moves that into `post-migration-improvements/claude-backend-improvements/` as entry 0018. Deleting the test before the entry exists loses the finding.

Verify the entry is on master before you delete anything, rather than assuming #87 merged.

## Worth saying in the PR

The natural future home for an executable check is the Go live suite in `backend/internal/http/`, which can assert the serialization directly without a second driver. That is a suggestion for the post-parity pass, not part of this ticket.

## Acceptance

- No `tests/parity/` directory; no `pg` in `package.json`.
- `bun install` clean; `bun run lint`, `bun test`, `bunx tsc --noEmit`, `bun run build` clean.
- Registry entry 0018 present on master.

Issue #86, opened 2026-08-09
Consolidate the per-domain mutation-error and path-param helpers


GonzaloSecades:
## Problem

The three frontend domains each grew their own copy of the same two helpers during #50:

```
features/accounts/api/account-mutation-error.ts
features/accounts/api/account-path-params.ts
features/categories/api/category-mutation-error.ts
features/categories/api/category-path-params.ts
features/transactions/api/transaction-mutation-error.ts
features/transactions/api/transaction-path-params.ts
```

Six files, two behaviors. The duplication was deliberate while the domains were separate in-flight PRs — reaching across lanes would have coupled them — and both copies carry a comment saying so. They have all merged now.

## Why it matters beyond tidiness

Both helpers encode a rule that is easy to get wrong and was in fact got wrong once each:

- **mutation-error** must read the API's *structured* message and fall back to the caller's wording. Using `error.message` surfaces `Failed to fetch accounts: 500 Internal Server Error` to a user. That defect shipped in the transactions lane and had to be fixed after the accounts lane had already found it.
- **path-params** must reject a missing id. `openapi-fetch` leaves the literal `{id}` in the URL when the value is undefined, so a request goes out to `/api/accounts/%7Bid%7D`.

Three copies means three chances to drift, and a fourth domain would copy whichever it found first.

## Scope

- Move both into `lib/api/` as domain-agnostic helpers — the resource name and fallback wording become parameters.
- Update the six call sites; delete the six files.
- Keep the existing tests, moved alongside. `account-mutation-error.test.ts` and `transaction-mutation-error.test.ts` both have coverage worth preserving, including the inherited-property case (`Object.hasOwn`, not `in`).

## Note

`lib/api/**` is the shared client layer. Confirm the move with Gonzalo before starting — this ticket is the request.

## Acceptance

- One implementation of each helper, no per-domain copies.
- `bun run lint`, `bun test`, `bunx tsc --noEmit`, `bun run build` clean.
- No behavior change: the same message precedence, the same guard.

GonzaloSecades:
Merged via PR #97.

Verification at merge:
- backend, frontend, and openapi checks: pass
- frontend confirmed frozen install, lint, `bunx tsc --noEmit`, `bun test`, and build
- unresolved review threads: 0

Master graphify artifacts were refreshed in commit `20b8d05`.

Issue #85, opened 2026-08-09
[Backend Migration 21b] Remove Drizzle, Neon and Clerk


GonzaloSecades:
Second half of #27. Target end state for the programme: **Next + Go, nothing else.** This ticket removes everything non-Go left on the data and auth side.

## Depends on two tickets

- **#84** — the Hono routes are the last consumers of the Drizzle schema and Clerk auth.
- **#89** — nine form/sheet components still import `InsertAccountSchema` / `InsertCategorySchema` / `InsertTransactionSchema` from `db/schema.ts`. `db/` cannot be deleted until those move to contract-derived types.

Starting before both have merged means deleting code that is still imported.

## Scope

**Drizzle / Neon**
- Delete `db/**` (`schema.ts`, `drizzle.ts`, `connection.ts`, `connection.test.ts`).
- Delete `drizzle/**` — the generated SQL migrations. `backend/migrations` (goose) is authoritative now.
- Delete `scripts/migrate.ts` and `scripts/seed.ts`.
- Update `scripts/dev-local.ts`, which invokes the migrate script at line 93.
- Remove `drizzle-orm`, `drizzle-zod`, `drizzle-kit` and the `db:generate` / `db:migrate` / `db:studio` / `db:seed` scripts from `package.json`.

**Clerk**
- Remove `@clerk/nextjs` and `@clerk/backend` from `package.json` (`@hono/clerk-auth` goes with #84).
- Remove the Clerk env vars from `.github/workflows/ci.yml` — the comment there says "the Clerk dependency disappears with #51/#27". This is that moment.
- Remove the whole Clerk block from `.env.example`: `NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY`, `CLERK_SECRET_KEY`, `CLERK_PUBLISHABLE_KEY`, `NEXT_PUBLIC_CLERK_SIGN_IN_URL`, `NEXT_PUBLIC_CLERK_SIGN_UP_URL`, and both `..._FALLBACK_REDIRECT_URL` entries.
- Review `ALLOW_DB_SEED` in `.env.example`; it exists only for the seed script this ticket deletes.

## Do NOT remove `pg`

`tests/parity/support/stacks.ts` still imports it. `pg` is removed in #90, which retires that harness. Check every package by usage rather than by name — `db/connection.ts` is not its only consumer.

## Acceptance

- `bun run lint`, `bun test`, `bunx tsc --noEmit`, `bun run build` all clean.
[...]

Issue #84, opened 2026-08-09
[Backend Migration 21a] Remove the legacy Hono API surface


GonzaloSecades:
First half of #27. Sequenced **before** #85, which removes Drizzle, Neon and Clerk.

Target end state for the whole programme: **Next + Go, nothing else.** This ticket removes the Hono half and the feature flag that selected between the two backends.

The frontend cutover is complete — every hook uses the generated OpenAPI client, no module imports `lib/hono`, and the Go API already serves everything. The Hono routes serve nothing.

## Scope

**Delete the Hono surface**
- `app/api/[[...route]]/**` — accounts, categories, summary, transactions, route.ts.
- `lib/hono.ts`.
- `hono`, `@hono/zod-validator` and `@hono/clerk-auth` from `package.json`.

**Remove `USE_GO_API` completely** (decided — do not keep an escape hatch)

With Hono gone there is no second backend to select, so the flag selects nothing. Every reference goes:

- `lib/go-api-rewrite.ts:82` — the flag branch, and `USE_GO_API` from the `RewriteEnv` type at line 15. The rewrite becomes unconditional.
- `lib/go-api-rewrite.test.ts` — lines 91, 92, 98, 117, 126, 136 all drive the function through the flag. Rewrite these tests around the unconditional behavior; keep the `GO_API_URL` validation coverage, which is still load-bearing.
- `.env.example:69`.
- `README.md` — the "Go API proxy configuration" section around lines 215-236, including the bullet that calls `USE_GO_API=false` "diagnostic only". Both bullets there reference the legacy routes and need rewriting, not just deleting.

`tests/parity/README.md` also mentions the flag, but that whole directory is deleted in #90 — leave it alone.

## Watch out

- `lib/go-api-rewrite.test.ts` has two pre-existing typecheck errors being fixed in #83. Land #83 first and pull master before editing, or you will both rewrite the same file.
- The rewrite currently sits in `beforeFiles` because that is the only phase that can override `app/api/[[...route]]`. Once that route is gone the *reason* disappears. Keeping `beforeFiles` is still correct — but confirm the behavior rather than reasoning about it, and say in the PR which phase you left it in and why.
- `tests/parity/**` reads the database through node-postgres and does not import Hono, so it survives this deletion and is retired separately in #90. Confirm it still runs here.

## Not in scope

[...]

PR #91, opened 2026-08-09
[Issue - #83] Gate CI on the test suite and typecheck

### LINKED ISSUE #83: Gate CI on the test suite and typecheck
## Problem

CI's `frontend` job runs only `bun run lint` and `bun run build`. It does **not** run `bun test` or `bunx tsc --noEmit`.

The 215 tests now on master therefore gate nothing. This is not theoretical: the refresh-envelope defect fixed in #80 shipped green through CI, and would have shipped green even with correct tests, because no job executed them.

With #27 (legacy teardown) next — a large deletion — the safety net should exist before it lands, not after.

## Scope

1. Add `bun test` and `bunx tsc --noEmit` to the `frontend` job in `.github/workflows/ci.yml`.
2. Fix the two pre-existing typecheck errors that would immediately fail that gate:
   - `lib/go-api-rewrite.test.ts:85` and `:129` — `'rewrites.beforeFiles' is possibly 'undefined'`.
3. Add `.worktrees/**` and `.claude/worktrees/**` to `globalIgnores` in `eslint.config.mjs`.

## Why the eslint ignore belongs here

The documented workflow puts worktrees at `.worktrees/<name>` inside the repo. ESLint ignores the root `.next/**` but not nested worktree builds, so `bun run lint` on master currently reports ~23,000 problems from sibling worktrees and takes minutes to finish. Master's own code is clean — every reported file is under `.worktrees/`. Local lint is unusable for both agents until this is ignored, which undermines the same verification this ticket is about.

## Acceptance

- CI fails on a failing test and on a type error (verify with a deliberate failure on a scratch branch, then revert).
- `bun run lint` on master completes quickly and reports zero problems with worktrees present.

GonzaloSecades:
Closes #83.

## What this does

Adds `bunx tsc --noEmit` and `bun test` to the `frontend` job. Typecheck runs
first: it is the cheapest of the three checks and reports a whole class of
failure at once, rather than one file per `next build` attempt.

Ignores `.worktrees/**` and `.claude/worktrees/**` in `eslint.config.mjs`.

## The ticket's premise was wrong, and it mattered

The ticket predicted two pre-existing typecheck errors. On a clean worktree with
`bun install --frozen-lockfile` there are **18**:

```
16 x error TS2307: Cannot find module 'bun:test' or its corresponding type declarations.
 2 x error TS18048: 'rewrites.beforeFiles' is possibly 'undefined'.
```

`bun-types` is only an *optional* peer of drizzle-orm, so a frozen-lockfile
install — which is what CI does — does not have it, and every one of the 16 test
files fails to resolve its import. Any environment that happened to have
`bun-types` on disk would see the 2 errors the ticket describes, which is
presumably where that number came from.

### The chain that follows from fixing it

1. `@types/bun` is the fix for the 16. It is now a devDependency.
2. It brings `bun-types` into `node_modules`, and
   `drizzle-orm/sqlite-core/columns/blob.d.ts` carries an unconditional
   `/// <reference types="bun-types" />`. That reference resolves the moment the
   package exists, so **Bun's globals load across the whole program** — this is
   not something `types` or `typeRoots` in tsconfig can prevent; I tried both.
3. Bun's `globals.d.ts` declares `namespace fetch { preconnect }`, so
   `typeof globalThis.fetch` now carries a static that a plain function cannot
   supply. `lib/api/authenticated-fetch.ts` annotated its return that way, and
   this broke **`next build`** as well as the new typecheck.

So this could not be landed green without touching that file. It now declares
the call signature it actually implements:

```ts
type FetchLike = (input: RequestInfo | URL, init?: RequestInit) => Promise<Response>;
```

**Heads-up for the #86 lane:** `lib/api/**` is Codex's during #86, and this
edits `lib/api/authenticated-fetch.ts`. #86's scope is the mutation-error and
path-param helpers and their six call sites, so it should not touch this file —
but flagging it rather than letting it surprise anyone. The drizzle reference
disappears when #85 removes drizzle-orm; the explicit type is correct either
way, so nothing here needs reverting afterwards.

### The two predicted errors

`rewrites.beforeFiles` is optional on Next's phased-rewrites type. Narrowed with
optional chaining rather than a non-null assertion, so a missing entry still
fails the assertion instead of throwing. Kept minimal because #84 rewrites this
file.

## The eslint ignore is load-bearing

Measured at the repo root with sibling worktrees present:

| | problems | time |
|---|---|---|
| before | 32,964 (796 errors, 32,168 warnings) | minutes |
| after | 0 | 41s |

Every reported file was under `.worktrees/`; master's own code is clean.

## Verification

All four run locally, on this branch, fresh (no cache reuse):

| check | result |
|---|---|
| `bunx tsc --noEmit` | clean, exit 0 |
| `bun test` | 207 pass, 8 skip, 0 fail (215 across 22 files) |
| `bun run build` | compiled, 12/12 static pages |

The 8 skips are `tests/parity/**`, which skip without Postgres and a running Go
API. That matches what its README documents for a no-services run, and that
directory is retired in #90.

Acceptance also requires proving the gate *fails*. Verified on a scratch branch
— see the comment below.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,7 +109,14 @@ jobs:
       NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY: ${{ secrets.NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY || 'pk_test_Y2xlcmsuZXhhbXBsZS5jb20k' }}
       CLERK_SECRET_KEY: ${{ secrets.CLERK_SECRET_KEY || 'sk_test_ci_placeholder_not_a_real_key' }}
       DATABASE_URL: postgres://nuchi:nuchi@localhost:5432/nuchi?sslmode=disable
-      NEXT_PUBLIC_API_URL: http://localhost:3000
+      # NEXT_PUBLIC_API_URL is deliberately unset. It is read in exactly one
+      # place — resolveApiBaseUrl in lib/api-base-url.ts — and only on the
+      # server; the browser always uses a relative base so requests stay
+      # same-origin and are proxied to Go. Setting it to an absolute origin made
+      # `bun test` disagree with a developer machine, where it is normally
+      # absent: lib/api/client.test.ts asserts API_BASE_URL is relative, and
+      # under bun there is no `window`, so the server branch is what runs.
+      # Unset is also what a real deployment behind the proxy looks like.
     steps:
       - uses: actions/checkout@v4
       - uses: oven-sh/setup-bun@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,4 +115,11 @@ jobs:
       - uses: oven-sh/setup-bun@v2
       - run: bun install --frozen-lockfile
       - run: bun run lint
+      # Until #83 this job ran only lint and build, so the frontend test suite
+      # gated nothing: the refresh-envelope defect fixed in #80 went through CI
+      # green because no job executed a test. Typecheck runs before the tests
+      # and the build because it is the cheapest of the three and localizes a
+      # whole class of failure that `next build` reports one file at a time.
+      - run: bunx tsc --noEmit
+      - run: bun test
       - run: bun run build
--- a/bun.lock
+++ b/bun.lock
@@ -52,6 +52,7 @@
       },
       "devDependencies": {
         "@tailwindcss/postcss": "^4",
+        "@types/bun": "^1.3.14",
         "@types/node": "^20",
         "@types/react": "^19",
         "@types/react-dom": "^19",
@@ -517,6 +518,8 @@
 
     "@tybys/wasm-util": ["@tybys/wasm-util@0.10.1", "", { "dependencies": { "tslib": "^2.4.0" } }, "sha512-9tTaPJLSiejZKx+Bmog4uSubteqTvFrVrURwkmHixBo0G4seD0zUxp98E1DzUBJxLQ3NPwXrGKDiVjwx/DpPsg=="],
 
+    "@types/bun": ["@types/bun@1.3.14", "", { "dependencies": { "bun-types": "1.3.14" } }, "sha512-h1hFqFVcvAvD9j9K7ZW7vd82aSA+rTdznZa+5bwvCwqSB1jmmfLcbIWhOLx1/+boy/xmjgCs/OMUL8hRJSmnPw=="],
+
     "@types/d3-array": ["@types/d3-array@3.2.2", "", {}, "sha512-hOLWVbm7uRza0BYXpIIW5pxfrKe0W+D5lrFiAEYR+pb6w3N2SwSMaJbXdUfSEv+dT4MfHBLtn5js0LAWaO6otw=="],
 
     "@types/d3-color": ["@types/d3-color@3.1.3", "", {}, "sha512-iO90scth9WAbmgv7ogoq57O9YpKmFBbmoEoCHDB2xMBY0+/KVrqAaCDyCE16dUspeOvIxFFRI+0sEtqDqy2b4A=="],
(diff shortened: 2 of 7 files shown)


GonzaloSecades:
## Gate verification

**The gate works, and it was not free — it caught four pre-existing failures on
its first run.**

### It fails on a failing test — proven, not assumed

Scratch branch with one deliberately failing test, opened as a draft PR so the
workflow ran. The frontend job:

```
5 Run bun run lint:        success
6 Run bunx tsc --noEmit:   success
7 Run bun test:            failure
8 Run bun run build:       skipped
```

Worth stating plainly: **that same branch would have been fully green under the
old lint-and-build-only workflow.** Scratch branch and PR are closed and
deleted; nothing from them is merged. The typecheck half is being proven the
same way on a second scratch branch — result posted below when it lands.

### What it caught immediately

The first real run of this PR was **red**, with 4 failures that pass on a
developer machine and fail on the runner. Neither was a new break; both were
already on master, invisible because nothing ran them.

**1. `lib/api/client.test.ts` — `API_BASE_URL` is relative**

Under bun there is no `window`, so `resolveApiBaseUrl` takes its *server*
branch and returns `NEXT_PUBLIC_API_URL`. The frontend job set that to
`http://localhost:3000`; a developer machine does not set it at all. So the
base came out absolute in CI and relative locally.

The variable is read in exactly one place — `lib/api-base-url.ts` — and only on
the server, and unset is what a real deployment behind the same-origin proxy
looks like. The job no longer sets it.

**2. `transaction-date.timezone.test.ts` — 3 failures**

This one is the more interesting find. The suite spawns a probe under several
`TZ` values, and **no single spelling of a zone works on both platforms**:

| | `GMT-3` | `Asia/Tokyo` |
| --- | --- | --- |
| Bun on Windows | resolves, UTC−3 | **silently ignored**, keeps host zone |
| Bun on Linux | **falls back**, keeps host zone | resolves, UTC+9 |

Measured, not inferred. On this machine (host at UTC−3) `TZ=Asia/Tokyo` reports
offset 180, not −540 — the IANA name did nothing and the host zone leaked
through. In CI, `GMT-3` and `GMT+9` both reported 0.

So all three cases ran at UTC on the runner, and the suite's own guard —
"really runs in that timezone" — correctly refused to pretend otherwise. That
guard is the reason this surfaced as a clean failure instead of three vacuous
passes.

Each zone now carries candidate spellings, and a candidate is accepted only
once the spawned process **reports the offset the case is about**. The offset
check was already there; making it do the selecting means a silent fallback can
never be mistaken for a working zone on either platform. When nothing works the
suite fails with every spelling it tried, rather than testing the host zone
three times over.

Both branches verified locally rather than reasoned about:

- forcing a bogus first candidate → it is rejected and the next one selected, suite passes
- making *every* candidate bogus → `no TZ spelling put the process at offset -540 … Tried: Bogus/NotAZone -> 180.`

Probe results are also memoized per zone now; the file was spawning the same
subprocesses nine times.

### Green run

| job | result |
| --- | --- |
| frontend | pass — all 8 steps, including `bunx tsc --noEmit` and `bun test` |
| backend | pass |
| openapi | pass |

`bun test` in CI: **207 pass, 8 skip, 0 fail** across 22 files — identical to
local. The 8 skips are `tests/parity/**`, which skip without Postgres and a
running Go API, as its README documents; that directory is retired in #90.

### Flagged, deliberately not done here

- `lib/api/client.test.ts` asserts a property of the *ambient environment*
  rather than of the code — it is named "in a browser-like environment" but
  never constructs one, and it silently depends on `NEXT_PUBLIC_API_URL` being
  absent. Removing it from the CI job fixes today's symptom, not the fragility.
  The honest fix is to test `resolveApiBaseUrl({ isBrowser: true })` directly,
  but `lib/api/**` is #86's lane, so I have left it alone rather than reach
  across.
- `.worktrees/` and `.claude/worktrees/` are untracked and **not** gitignored,
  so they show up in `git status` and could be committed by accident. `.gitignore`
  currently has an unrelated uncommitted local edit, so I have not touched it.

### Note on the ownership line I crossed

This edits `lib/api/authenticated-fetch.ts`, which sits in #86's lane. It was
not optional: adding `@types/bun` is what makes `bun:test` typecheck at all, and
it breaks `next build` — the *pre-existing* gate — until that annotation is
fixed. #86's scope is the mutation-error and path-param helpers and their six
call sites, so it should not touch this file, but flagging it explicitly rather
than letting it surprise anyone.

GonzaloSecades:
## Gate verification — the typecheck half

Both halves of the acceptance criterion are now proven on the runner rather
than assumed from the YAML parsing.

Scratch branch with a single deliberate type error in a file nothing imports:

```
5 Run bun run lint:        success
6 Run bunx tsc --noEmit:   failure
7 Run bun test:            skipped
8 Run bun run build:       skipped
```

```
scratch-typecheck-proof.ts(3,14): error TS2322: Type 'string' is not assignable to type 'number'.
##[error]Process completed with exit code 2.
```

Scratch branch and PR (#94) closed and deleted; nothing from them is merged.

### Summary of the two proofs

| deliberate breakage | job fails at | steps skipped after |
| --- | --- | --- |
| failing test | step 7 `bun test` | `bun run build` |
| type error | step 6 `bunx tsc --noEmit` | `bun test`, `bun run build` |

The first of those is the one that matters most: **that branch would have been
fully green under the old lint-and-build-only workflow**, which is exactly how
the #80 refresh-envelope defect shipped.

My first attempt at the typecheck proof died on an unrelated transient
`bun install` failure (`error: Fail extracting tarball for "next"`), so it
proved nothing and was re-run rather than written up as a pass.

PR #93, opened 2026-08-09
[Issue - #87] Record the transaction date serialization divergence

### LINKED ISSUE #87: Record the transaction date serialization divergence
## The finding

Found by the differential parity harness (#82), the first thing it was pointed at. Not yet recorded anywhere.

`transactions.date` is `timestamp without time zone`. Both stacks read the same bytes and disagree about what timezone those bytes are in:

| | value |
| --- | --- |
| stored | `2026-08-07 00:00:00` |
| node-postgres (legacy Hono) | `2026-08-07T03:00:00.000Z` |
| Go | `2026-08-07T00:00:00Z` |

A browser renders those as **different calendar days** anywhere west of Greenwich — every transaction date a day early, including for Argentina, which is the app's own market and its default currency.

## Status

The frontend has been made resilient to it: `features/transactions/api/transaction-date.ts` converts explicitly at each edge, so display and the edit round trip are correct regardless of which serialization arrives. **The backend serialization itself is unchanged.**

Registry entry 0014 covers the *filter* side of the same underlying modelling problem, and `docs/api/transactions.md` notes under "Known gaps" that `date` is a `timestamp` modelling what is really a calendar date. The **response** side is not recorded.

## Scope

1. Write `post-migration-improvements/claude-backend-improvements/0018-date-response-serialized-as-utc.md` using the template — ticket link, how it was migrated, why, the concern, the proposed improvement. Cross-reference 0014 and the "Known gaps" section.
2. Add it to the registry index.
3. Add a "Divergences" row to `docs/api/transactions.md` pointing at the new entry, matching how 0006/0014/0015/0016 are listed there.

## Deliberately not in scope

Changing the backend. The proposed fix — storing and serializing a real `date` rather than a naive `timestamp` — is a schema and contract change, and belongs to the post-parity hardening pass. This ticket only makes sure the finding survives in writing.

## Why now

The parity harness asserts this divergence is still present and classifies it as `open` rather than expected, so the test fails if someone reclassifies it. That test is currently the only record. A registry entry is the durable one.

GonzaloSecades:
Closes #87.

Documentation only. **No backend change** — the fix is a schema and contract
change and belongs to the post-parity pass.

## What lands

- `post-migration-improvements/claude-backend-improvements/0018-date-response-serialized-as-utc.md`
- The registry index row in `post-migration-improvements/README.md`
- A "Divergences" row in `docs/api/transactions.md`, matching how 0006/0014/0015/0016 are listed there

## Why it has to be now

The parity harness holds the only executable record of this divergence, and #90
deletes that harness. This has to be on master before #90 starts or the finding
goes with it.

## Two things I verified rather than restated

The ticket's table gives Go's output as `2026-08-07T00:00:00Z`. That is right,
and it is worth noticing that it is **not** what the fixtures froze — those
carry `2026-06-30T00:00:00.000Z`, node-postgres's millisecond form. So even at
UTC, where the two stacks agree on the instant, the serialized strings differ.
Confirmed by marshalling a `time.Time` through `encoding/json` the same way the
generated type does, rather than inferring it:

```
{"date":"2026-08-07T00:00:00Z"}
```

Nothing observable in the app turns on it — both parse to the same instant in
JavaScript — but a byte comparison against the fixtures would, so the entry
records it.

Second: the contract already contains the modelling error in plain sight. A
write takes `DateString` (`format: date`, `2026-06-30`) and the matching read
returns `format: date-time` with example `2026-06-30T00:00:00.000Z`. A calendar
date goes in and an instant comes out. That framing is in the entry because it
makes the proposed fix obvious — change the column and the response schema
together, rather than patching the serializer.

## Cross-references

Entry **0001** (the column type), entry **0014** (the filter side of the same
modelling problem), and the "Known gaps" note in `docs/api/transactions.md`. The
entry argues all three should land together in the post-parity pass rather than
being patched at separate layers.

The entry also notes that the current frontend survives the eventual fix
unchanged — `calendarDateFromApi` already passes a plain `yyyy-MM-dd` through —
while flagging that this should be confirmed, not assumed, when the work is
picked up.

## Verification

`prettier --check` clean on the new entry, matching its siblings. Note that
`post-migration-improvements/README.md` and `docs/api/transactions.md` already
fail `prettier --check` **on master** — markdown formatting is not enforced by
any CI job — so I left their existing formatting alone rather than burying a
one-row change under a whole-file reformat.

No code changed, so lint/test/build are unaffected.

### DIFF
--- a/docs/api/transactions.md
+++ b/docs/api/transactions.md
@@ -363,6 +363,7 @@ Recorded in `post-migration-improvements/claude-backend-improvements/`:
 | 0014 | Date filters parsed in UTC rather than the host timezone |
 | 0015 | Out-of-range amount returns 400 rather than legacy's 500 |
 | 0016 | Bulk byte limits enforced against the stream, not just `Content-Length` |
+| 0018 | `date` responses serialize the naive timestamp as UTC; legacy's driver read it as host-local |
 
 Also intentional, driven by the contract rather than parity: `currency` required
 (legacy has no such field), empty query parameters rejected, empty `categoryId`
--- a/post-migration-improvements/README.md
+++ b/post-migration-improvements/README.md
@@ -108,3 +108,4 @@ gates. Numbering stays global and sequential within
 | 0015 | [Out-of-range amounts return 400, where legacy returned 500](claude-backend-improvements/0015-out-of-range-amount-returns-400.md) | api | low |
 | 0016 | [Bulk body limits are enforced against the stream, not just Content-Length](claude-backend-improvements/0016-bulk-body-limit-enforced-against-the-stream.md) | http | low |
 | 0017 | [The summary category chart silently excludes uncategorized spending](claude-backend-improvements/0017-summary-category-chart-excludes-uncategorized.md) | api/product | medium |
+| 0018 | [A transaction's date is serialized as a UTC instant, not a calendar date](claude-backend-improvements/0018-date-response-serialized-as-utc.md) | api/schema | medium |
--- a/post-migration-improvements/claude-backend-improvements/0018-date-response-serialized-as-utc.md
+++ b/post-migration-improvements/claude-backend-improvements/0018-date-response-serialized-as-utc.md
@@ -25,6 +25,10 @@ Writes are unaffected: the contract's `DateString` (`format: date`, pattern
 `^\d{4}-\d{2}-\d{2}$`) is parsed to a naive `time.Time` and stored as midnight.
 Only the response side is at issue here.
 
+That midnight is a property of the current write path, not of the table. The
+column carries no `CHECK`, and the legacy path stored arbitrary instants — which
+is why the proposed fix below begins with an audit rather than an `ALTER TABLE`.
+
 ## Why it was done this way
 
 Parity froze the column type. Entry 0001 records why the `timestamp` survived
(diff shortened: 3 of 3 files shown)


GonzaloSecades:
## Addressed: audit before truncating (P2)

**The finding is correct on all four evidence points**, verified against the
source rather than accepted:

| claim | verified |
| --- | --- |
| `z.coerce.date()` on the legacy field | `db/schema.ts:116`, inside the `InsertTransactionSchema` override |
| Hono writes the parsed `Date` directly | the create handler spreads `...values` straight into `db.insert(transactions)` |
| no `CHECK` requiring midnight | `date timestamp NOT NULL` in `00002_finance_base.sql`; the only `CHECK`s in `backend/migrations/` are RLS policy clauses |
| the legacy schema accepts `2026-08-07T15:45:00Z` | follows from `z.coerce.date()`, which coerces any parseable instant |

So "data-preserving since every stored value is already midnight" asserted an
unverified property of production data, in a document whose whole purpose is to
seed a future migration ticket. That is a defect in migration guidance, not a
wording nit.

### The risk is worse than "the time component is discarded"

While correcting this I chased down *why* a non-midnight row would exist, and
the mechanism makes the second half of your risk note concrete. node-postgres
serializes a JS `Date` into a zone-less `timestamp` using the **Node process's
local wall clock** — `dateToString` in `pg/lib/utils.js` reads `getFullYear()`
and `getTimezoneOffset()`, and Postgres discards the trailing offset for a
zone-less column. The stored wall time is therefore the chosen instant rendered
in the *legacy host's* zone, and which calendar day it lands on depends on the
gap between that zone and the user's:

| user picks | user zone | legacy host | stored | `date::date` |
| --- | --- | --- | --- | --- |
| 2026-08-07 | UTC−3 | UTC−3 | `2026-08-07 00:00:00` | 2026-08-07 ✅ |
| 2026-08-07 | UTC−3 | UTC | `2026-08-07 03:00:00` | 2026-08-07 ✅ |
| 2026-08-07 | UTC+9 | UTC | `2026-08-06 15:00:00` | 2026-08-06 ❌ |

In the third row the time component is not noise — it is the only thing
identifying the right day, and `date::date` destroys that evidence in the same
statement that gets the answer wrong. Silent and irreversible.

### What changed

`Proposed improvement` is now **audit → normalize → convert**:

1. **Audit**, with your suggested query, plus why midnight was never enforced
   (no `CHECK`; `z.coerce.date()`; the spread into the insert).
2. **Decide an explicit normalization policy** for whatever the audit finds,
   including which zone the legacy values are to be read in — with the mechanism
   and the table above, and an instruction to confirm it against deployment
   history rather than assume it.
3. **Then** convert with `ALTER TABLE ... USING date::date`, qualified as
   lossless *only* in the state steps 1 and 2 establish, and explicitly called
   out as a silent irreversible data change if run first.

Steps 4 and 5 (contract change, regenerate/re-cut fixtures) are unchanged; the
"Step 2 is a breaking response change" cross-reference was renumbered to Step 4.

I also added a qualifier to **How it was migrated**, where the entry says the Go
write path stores midnight — that is true of the write path but not of the
table, and without the qualifier a reader carries the assumption into the
section below.

### Verification

```
bunx prettier --check post-migration-improvements/claude-backend-improvements/0018-date-response-serialized-as-utc.md
Checking formatting...
All matched files use Prettier code style!
```

Clean under both the repo-pinned prettier and the `bunx` resolution.

Diff remains **documentation-only** — three `.md` files, zero non-`.md` paths:

```
docs/api/transactions.md                              |   1 +
post-migration-improvements/README.md                 |   1 +
.../0018-date-response-serialized-as-utc.md           | 173 +++++++++++++
```

No backend, migration, contract, or runtime change. Pushed to
`claude/87-date-serialization-entry` as `6f0ffd7`; branch not merged, not
deleted.

PR #97, opened 2026-08-09
[Issue - #86] Consolidate per-domain API helpers

### LINKED ISSUE #86: Consolidate the per-domain mutation-error and path-param helpers

GonzaloSecades:
Closes #86.

## What changed

- Replaced six per-domain mutation-error and path-parameter helpers with one shared implementation of each under `lib/api/`.
- Updated the account, category, and transaction hooks to use the shared helpers.
- Moved and consolidated regression coverage beside the shared code, including the inherited-property case requiring `Object.hasOwn`.

## Behavior preserved

- Toasts read the API's structured error message and fall back to caller wording; the generic technical `error.message` is never surfaced.
- Missing IDs throw before `openapi-fetch` can send a literal `{id}` placeholder.
- Account, category, and transaction missing-ID messages remain unchanged.

## Validation

- `bun run lint` — pass
- `bun test` — pass; parity tests skip without external services
- `bun run build` — pass
- Targeted shared-helper tests — 14 pass
- `bunx tsc --noEmit` — only the pre-existing missing `bun:test` declarations and two `lib/go-api-rewrite.test.ts` narrowing errors fixed by #83

`graphify update .` completed locally; generated `graphify-out/` changes are intentionally excluded.

## Merge order

Draft until #83 lands. #83 must merge first so the full test/typecheck CI gate protects this teardown work.


GonzaloSecades:
Addressed the scoped follow-up:

- removed the unused `overrides` parameter, API-code read, and override branch from `mutationErrorMessage`
- removed `code` from the local envelope type
- deleted exactly the five override-related tests and retained the five structured-message/fallback tests
- preserved the explanatory doc comment verbatim
- left all call sites and path-parameter helpers untouched

Verification:
- `bun run lint` — pass
- `bunx tsc --noEmit` — pre-#83 baseline only: 15 missing `bun:test` declarations and 2 known `rewrites.beforeFiles` errors; 0 other errors
- `bun test` — 199 pass, 8 parity skips, 0 fail; 207 total tests across 21 files
- `bun run build` — pass with the documented Clerk placeholders

The supplied count also said the branch had 212 total tests before removing five; Bun's 199 pass + 8 skip = 207 total is the consistent result. No tests were added to manufacture the contradictory 207-pass/8-skip count.

Issue #36, opened 2026-07-01
[Backend Migration 02] Define shared API error and auth contract


GonzaloSecades:
Parent spec: #18.
Related epic: #29.
Spec source: `docs/specs/18-go-backend-replacement/spec.md`.

Graphify note:
- Use the current graph output to confirm impacted endpoints, flows, and business logic before implementing changes.
- Refresh the graphify artifacts before handing this ticket in or before starting another migration ticket.

Define the shared contract details that every Go endpoint and frontend client hook must use.

Acceptance criteria:
- Structured API error format is defined in OpenAPI.
- Bearer access-token auth is represented in OpenAPI.
- Refresh-token cookie behavior is represented in OpenAPI.
- Unauthorized and expired-token behavior is documented.
- App resource success responses preserve `{ data: ... }` where practical.
- Auth endpoint response shapes are documented separately from app resource envelopes.

Verification:
- OpenAPI validation command passes.

GonzaloSecades:
Completed in dbc2efd.\n\nImplemented the shared OpenAPI API error/auth contract for Backend Migration 02:\n- structured ApiErrorResponse/ApiError schemas\n- Bearer JWT access-token security scheme\n- HttpOnly refresh-token cookie scheme and Set-Cookie headers\n- auth endpoint request/response shapes kept separate from app resource envelopes\n- unauthorized and expired-token examples\n- app resource { data: ... } envelope guidance\n\nGraphify handoff:\n- installed the project Graphify Codex skill/instructions\n- ran graphify update . after the contract changes\n- refreshed graphify-out artifacts built from fbcbc66b\n- Graphify reported AST-only/no LLM cost plus extraction warnings: 13 zero-node files and 17 missing source_file issues\n\nVerification:\n- bun run openapi:validate passed\n- bun run lint passed\n- cd backend && go test ./... passed\n- Docker Desktop services postgres and mailpit were verified healthy locally\n\nNote: bun run build still requires real Clerk values in .env.local; current placeholders intentionally fail Clerk prerender validation.

PR #98, opened 2026-08-09
[Issue - #84] Remove the legacy Hono API surface and USE_GO_API

### LINKED ISSUE #84: [Backend Migration 21a] Remove the legacy Hono API surface

GonzaloSecades:
Closes #84. First half of #27.

## What this removes

- `app/api/[[...route]]/**` — accounts, categories, summary, transactions, `route.ts`. `app/api/` is now empty and gone.
- `lib/hono.ts` — the typed RPC client. No module imported it.
- `hono`, `@hono/zod-validator`, `@hono/clerk-auth` from `package.json`.
- `USE_GO_API`, completely — from `lib/go-api-rewrite.ts`, its tests, `.env.example` and `README.md`. With Hono gone there is no second backend to select.

`tests/parity/README.md` still mentions the flag and was deliberately left alone; that whole directory is retired in #90.

## Which rewrite phase, and why

**Left in `beforeFiles`.** The original justification — that `beforeFiles` is the only phase that can override `app/api/[[...route]]` — died with the route. `afterFiles` would fire correctly now. `beforeFiles` is still the right choice for a different reason: it keeps the proxy authoritative, so a Next route added under `app/api/` later cannot silently shadow the backend. The code comment and the test comment both say this now instead of citing the deleted route.

Confirmed by behavior, not by reasoning. With the route deleted, Go on :8080 and `next dev` on :3000:

```
$ curl -i http://localhost:3000/api/health
HTTP/1.1 200 OK
{"service":"nuchi-api","status":"ok","time":"2026-08-09T21:24:20Z"}

$ curl -i http://localhost:3000/api/accounts
HTTP/1.1 401 Unauthorized
{"error":{"code":"UNAUTHORIZED","message":"Authentication required."}}
```

The 401 is the tell: that structured shape is the Go contract. Legacy Hono returned plain `{"error":"Unauthorized"}`. `USE_GO_API` was absent from the environment entirely for both requests.

## Tests

`lib/go-api-rewrite.test.ts` went from 15 tests to 14. All nine `normalizeGoApiUrl` cases are untouched and still load-bearing — origin-only rejection, trailing-slash normalization, protocol rejection, embedded-credential rejection, the named-offending-value case. The one test that disappeared is `allows an explicit non-true value to disable the rewrite`, which tested the flag; the remaining five `goApiRewrites` tests now drive the function through an empty or origin-only environment.

## Verification

All run in this worktree on this branch.

| check | result |
| --- | --- |
| `bunx tsc --noEmit` | clean |
| `bun test` | **198 pass, 8 skip, 0 fail** — 206 tests across 21 files |
| `bun run build` | passes; route list no longer contains `/api/[[...route]]` |

The 8 skips are `tests/parity/**`, unchanged by this PR — the harness does not import Hono and still runs. It is retired in #90, which is where the skip count reaches 0.

## Not in scope

`db/**`, `drizzle/**`, `@clerk/*`, `scripts/migrate.ts`, `scripts/seed.ts` → #85. `tests/parity/**` and `pg` → #90.

`README.md` statements that described the deleted files as the current backend were rewritten, since leaving them would point readers at paths this PR removes. Its remaining Drizzle/Clerk wording is accurate until #85 and was left for that PR. `AGENTS.md`, `CLAUDE.md`, `docs/CODEX_CONTEXT.md` and `docs/specs/**` also describe the Hono stack and were **not** touched — they need a decision of their own and are raised separately.


GonzaloSecades:
Review findings — changes requested before merge:

1. **[P2] Update the active rewrite comment in `next.config.ts`.** Lines 11–13 still say the proxy is “on by default” and direct readers to an “explicit diagnostic escape hatch” in `lib/go-api-rewrite.ts`. This PR removes that escape hatch and makes the rewrite unconditional, so the configuration comment becomes false immediately after merge. Please describe the unconditional proxy and the new authoritative-`beforeFiles` rationale instead.

2. **[P2] Do not leave the active repository guidance pointing at deleted Hono code.** `AGENTS.md` still declares Hono and `lib/hono.ts` as the stack and tells agents to extend typed Hono routes; `CLAUDE.md` still calls the Hono routes existing legacy code; and `docs/CODEX_CONTEXT.md` describes the current request flow through `app/api/[[...route]]` and `lib/hono.ts`. Those are operational/current-state documents, unlike the historical spec, parity fixtures, improvement entries, and `tests/parity/README.md`, which should retain their Hono references. The PR body says these were “raised separately,” but there is no linked follow-up issue. Please update the active guidance here, or link a concrete blocking follow-up that lands before #85.

Non-blocking documentation cleanup: the edited README still says the next migration issue is #29 (lines 20 and 330), while this PR makes #85 the next ticket. Since this PR is already rewriting Migration Status and the roadmap, it would be clearer to correct those two stale statements in the same pass.

The implementation checks are otherwise clean: no surviving runtime imports of the deleted route/client; rewrite tests 14/14 pass; full suite 198 pass / 8 parity skips / 0 fail; lint, typecheck, and build pass; the build route list contains no `/api/[[...route]]`.

GonzaloSecades:
## Review findings addressed — all three verified before changing anything

### 1. [P2] `next.config.ts` comment — correct, and worse than stated

The comment did not just become stale; it pointed readers at a file section
this PR deletes. `next.config.ts` was **not in the original diff**, which is
exactly why it slipped past.

Rewritten to describe the unconditional proxy, and to carry the `beforeFiles`
rationale in short form: `afterFiles` would fire too now that no filesystem
route matches `/api/*`, but a future Next route under `app/api/` would silently
shadow the backend from that phase, so `beforeFiles` keeps the proxy
authoritative. The long-form argument stays in `lib/go-api-rewrite.ts`.

### 2. [P2] Operational docs pointing at deleted code — fixed here, not deferred

Your distinction between operational and historical documents is the right cut,
and I applied exactly that line.

**Updated** (current-state docs an agent reads before working):

| file | was |
| --- | --- |
| `AGENTS.md` | declared Hono as the stack, told agents to "prefer extending typed Hono routes" |
| `CLAUDE.md` | called the Hono routes existing legacy reference material |
| `docs/CODEX_CONTEXT.md` | described the whole request flow through `app/api/[[...route]]` and `lib/hono.ts` |

**Deliberately not updated:** the Drizzle and Clerk references in those same
files. They are **still true** until #85 removes them, and rewriting them now
would make the docs wrong in the other direction for the window between this
merge and #85's. Each file names #85 as where the rest goes. That is the
follow-up you asked to be linked, and it is the next ticket rather than a
speculative one.

**Left alone on purpose:** the migration spec, `api-parity-fixtures.md`, the
improvement registry, and `tests/parity/README.md`. `CLAUDE.md` now says so
explicitly, so the next person does not "fix" them.

Two corrections beyond the Hono swap, both in a file already being edited:

- `AGENTS.md` "Current Risk Areas" listed Hono-implementation risks the Go
  migration already resolved — ownership predicates plus RLS, strict date and
  body validation, rate limiting. Two genuinely open frontend items kept.
- `AGENTS.md` "Verify" listed a weaker check set than CI enforces. It was
  missing `bunx tsc --noEmit` and `bun test`, which #83 made gating — so the
  guidance was telling agents a green local run meant something it did not.

### 3. [non-blocking] README `#29`

Taken, and it was in three places rather than two — the Migration Status
pointer, the deferred-generation note at line 273 (which claimed generation was
deferred "until #29 fills the full resource contract"), and the roadmap tail.
All corrected to point at #85 then #90. The completed-issues list is now marked
as covering the early tickets rather than implying it is exhaustive, since it
stops at #36 while 28 tickets are done.

## Verification

| check | result |
| --- | --- |
| `bunx tsc --noEmit` | clean |
| `bun test` | 198 pass, 8 skip, 0 fail — 206 across 21 files |
| `bun run build` | passes; **route list contains no `/api/[[...route]]`** |

Residual `hono` / `USE_GO_API` matches outside `backend/`, `graphify-out/`,
`.next/` and `bun.lock` are all accounted for: the historical records, the
deliberate "this was removed in #84" sentences, and two false positives — the
word "hon**our**s" in `lib/auth/errors.test.ts` and "hon**or**" in a Codex skill
file.

Copilot could not review this PR — its response on both attempts was a quota
limit, not a clean pass. Noting that rather than reporting it as reviewed.

Issue #89, opened 2026-08-09
Replace the drizzle-zod form schemas with contract-derived types


GonzaloSecades:
**Blocks #85.** `db/schema.ts` cannot be deleted until this lands.

## Problem

The forms and sheets still derive their validation schemas from Drizzle via `drizzle-zod`:

```
features/accounts/components/account-form.tsx
features/accounts/components/edit-account-sheet.tsx
features/accounts/components/new-account-sheet.tsx
features/categories/components/category-form.tsx
features/categories/components/edit-category-sheet.tsx
features/categories/components/new-category-sheet.tsx
features/transactions/components/edit-transaction-sheet.tsx
features/transactions/components/new-transaction-sheet.tsx
features/transactions/components/transaction-form.tsx
```

Nine files importing `InsertAccountSchema`, `InsertCategorySchema` or `InsertTransactionSchema` from `db/schema.ts`.

This is the last real dependency on Drizzle in the app, and it is a correctness problem as well as a teardown blocker: the forms validate against the **database** shape while the API validates against the **contract**. Those are not the same. `currency` is required by the contract and absent from the Drizzle schema entirely; `date` is a calendar date on the wire and a timestamp in the table.

## Scope

Replace them with zod schemas derived from — or checked against — `lib/api/generated/schema.d.ts` (`AccountInput`, `CategoryInput`, `TransactionInput`).

Keep the form-level shapes the components already use. `transaction-form.tsx` has its own `formSchema` where `amount` is a string and `date` a `Date`, converted on submit; that boundary is deliberate and should survive. What changes is where the *API* shape comes from.

## Watch out

- `features/transactions/api/transaction-payload.ts` already adapts form values to `TransactionInput`, including `currency` and the calendar-date conversion. Reuse it rather than duplicating the mapping.
- The amount bound in `transaction-form.tsx` (`isSafeMiliunitAmount`) is deliberate and unrelated to Drizzle — keep it.
- `scripts/seed.ts` also imports `db/schema.ts`, but that script is deleted in #85, not here.

## Acceptance

- No component imports from `db/**`.
- `bun run lint`, `bun test`, `bunx tsc --noEmit`, `bun run build` clean.
- Creating and editing an account, a category and a transaction still work end to end against the Go API.

GonzaloSecades:
Merged via PR #95.

Verification at merge:
- backend, frontend, and openapi checks: pass
[...]

PR #99, opened 2026-08-09
[Issue - #85] Remove Drizzle, Neon and Clerk

### LINKED ISSUE #85: [Backend Migration 21b] Remove Drizzle, Neon and Clerk

GonzaloSecades:
Closes #85. Second half of #27 — after this, only #90 remains.

## What lands

**Drizzle / Neon:** `db/**`, `drizzle/**`, `scripts/migrate.ts`,
`scripts/seed.ts`, and the `drizzle-orm` / `drizzle-zod` / `drizzle-kit`
packages with their four `db:*` scripts.

**Clerk:** `@clerk/nextjs` and `@clerk/backend`, the `.env.example` block, and
the CI placeholder keys.

## Three things the ticket did not list, each found by checking usage not names

- **`drizzle.config.ts`** was still at the repo root, importing `drizzle-kit`
  and pointing at `./db/schema.ts` and `./drizzle`. After the listed deletions
  it referenced three things that no longer exist.
- **`dev:neon`** was still a script. Neon was never a package here — just a
  hostname check in `db/connection.ts` and that script name — so it would have
  been the last thing in the repo named after a departed provider, and it would
  still have *worked*, which is why nothing would flag it.
- **`dotenv`** became orphaned by this PR's own deletions. Its only consumers
  were `drizzle.config.ts`, `scripts/migrate.ts` and `scripts/seed.ts`.

## `bun dev` would have broken silently

`scripts/dev-local.ts:93` ran `bun ./scripts/migrate.ts`. **No automated check
covers the dev script** — lint, typecheck, tests and build all pass with it
broken, and CI never runs it.

It no longer runs migrations at all. Goose owns those from `backend/`, and the
frontend does not connect to Postgres any more.

That retired **`docker-compose.dev.yml`**, whose only consumer was that script.
Worth stating plainly: it ran a separate `nuchi_dev` database on port 54329
whose bootstrap superuser *was* the application role — exactly the layout
`docker-compose.yml` documents as unsafe, since superusers bypass row level
security outright, `FORCE` included, so the RLS policies would silently not
apply to anything using it. `dev-local.ts` now brings up the real `postgres`
and `mailpit` services.

## `.env.example` now meets #27's acceptance criteria

#27 requires it to document "the new backend, database, JWT, and mail settings".
The JWT ones were **missing**, so that criterion was not met on master:

| variable | why it is now there |
| --- | --- |
| `AUTH_JWT_SECRET` | required by the Go API, no default, startup exits if unset or under 32 bytes. Empty placeholder with `openssl rand -base64 48` as the hint — never a committed value. |
| `AUTH_COOKIE_SECURE` | defaults to `false`; noted that it must be `true` anywhere deployed, since it needs HTTPS. |

Removed: the Clerk block, `ALLOW_DB_SEED` (existed only for the deleted seed
script), and `HF_TOKEN` — no TypeScript, Go, or compose file read it.
`MAILPIT_WEB_URL` was kept: nothing reads it either, but it documents the
Mailpit UI address a developer actually uses.

## Documentation follow-through

#84 deliberately updated only the Hono references and left Drizzle and Clerk
alone, because those were still true until now. This is the other half —
`AGENTS.md`, `CLAUDE.md`, `README.md`. Historical references in the migration
spec, the fixtures, the improvement registry and `tests/parity/README.md` stay
untouched by design.

## `pg` deliberately kept

`tests/parity/support/stacks.ts` still imports it. #90 removes both together.

## Verification

| check | result |
| --- | --- |
| `bunx tsc --noEmit` | clean |
| `bun test` | 193 pass, 8 skip, 0 fail — 201 across 20 files |
| `bun run build` | **passes with no Clerk keys in the environment at all** |

The test delta is exactly `db/connection.test.ts` (5 tests, 1 file). The 8 skips
are `tests/parity/**`, retired in #90.

The build result is the acceptance criterion, so it was run with
`env -u NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY -u CLERK_SECRET_KEY` rather than by
just omitting them from a shell that might have had them.

Lockfile carries **zero** `@clerk`, `drizzle`, `hono` or `dotenv` entries. I
verified this against a **wiped** `node_modules` rather than an incrementally
pruned one — `bun remove` left stale `@clerk/clerk-react` and `@clerk/shared`
directories behind, which could have masked a surviving import locally.

## Flagged, not done

`@paralleldrive/cuid2` is now unused by any JavaScript — its consumers were the
Hono routes deleted in #84, and the Go side has its own implementation in
`backend/internal/id/cuid2.go`. It is a #84 leftover rather than something this
ticket orphaned, so I left it. Natural to fold into #90 alongside `pg`.


scripts/dev-local.ts
```diff
@@ -80,17 +97,8 @@ async function waitForPostgres() {
...
-  await run([
-    'docker',
-    'compose',
-    '-f',
-    COMPOSE_FILE,
-    'up',
-    '-d',
-    SERVICE_NAME,
-  ]);
+  await run(['docker', 'compose', 'up', '-d', ...COMPOSE_SERVICES]);
```
GonzaloSecades, line 116: [P2] Preserve a migration step in the fresh local setup path. On a new Compose volume, `docker/postgres/init/01-app-role.sql` creates only the application role and `citext`; it does not create `accounts`, `categories`, or `transactions`. The Go API does not migrate on startup and `VerifyRLSActive` immediately checks those three tables, so the PR's current `docker compose up` → `go run ./cmd/api` → `bun dev` instructions cannot bootstrap a clean checkout. Please either run the goose migrations from this root orchestration step after readiness, or add an explicit `goose -dir migrations postgres "$DATABASE_URL" up` step before API startup in the canonical "Running Next and Go Together" flow. The important part is that deleting the Drizzle runner must not silently delete the only migration step from fresh local development.

AGENTS.md
```diff
@@ -5,19 +5,17 @@
 - Next.js App Router (frontend only — there is no Next API layer)
 - Go API (chi, pgxpool, sqlc, goose) over PostgreSQL, reached same-origin
   through the `/api/*` rewrite in `next.config.ts`
-- Owned JWT auth (`lib/auth/`), not Clerk
+- Owned JWT auth (`lib/auth/`)
 - TanStack Query + the generated OpenAPI client in `lib/api/`
 - Bun package manager
-- Drizzle ORM and the `@clerk/*` packages are still installed but unused by any
-  code path; they are removed in #85
+- No ORM and no Clerk: both were removed in #85. The Go API owns all database
```
GonzaloSecades, line 11: [P2] The authoritative guide is still internally inconsistent: the Env section at lines 20–25 continues to instruct future agents to configure `NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY` and `CLERK_SECRET_KEY`, even though this new line, `.env.example`, and CI all say Clerk is gone. Please remove those stale variables and replace them with the relevant Go/runtime settings (at minimum the required `AUTH_JWT_SECRET`, plus the API/proxy settings you want surfaced).

GonzaloSecades (commented):
Review completed against issue #85, the dependency chain (#84 + #89), and the current head. The package/deletion boundary and retained parity `pg` dependency look correct, and the full verification suite is green. I found two changes needed before merge; details are inline.

GonzaloSecades:
## Both findings addressed — and both were right

### 1. [P2] Fresh local setup could not bootstrap

Verified against the source rather than accepted:

| claim | evidence |
| --- | --- |
| the init script creates no tables | `docker/postgres/init/01-app-role.sql` creates the `nuchi` role, hands it the `public` schema, and installs `citext`. That is all. |
| the API does not migrate on startup | no goose or migration call anywhere in `backend/cmd/api/main.go` |
| `VerifyRLSActive` checks those three tables immediately | `ownedTablesRequiringRLS = []string{"accounts", "categories", "transactions"}` (`backend/internal/db/rls.go:86`) |
| it refuses to serve | `main.go:44-48` — logs `refusing to start`, `pool.Close()`, `os.Exit(1)` |

And the failure is worse than "RLS is not active": `VerifyRLSActive` runs
`SELECT row_security_active($1::text)`, which **errors outright** on a table
that does not exist. So a clean checkout following the documented flow got a
startup error naming row level security, when the actual problem was an empty
schema — about as misleading as a first-run error can be.

**Fixed** by adding an explicit goose step to the canonical "Running Next and Go
Together" flow, between Postgres and the API, with the pinned install command
and the reason it is not optional. `scripts/dev-local.ts` now points at it as
well, so deleting the Drizzle runner does not read as "migrations do not
matter".

I took the documentation route rather than invoking goose from `dev-local.ts`.
Running it there would put the backend's schema ownership — and a Go toolchain
dependency — inside the frontend dev server, which is the coupling this PR is
removing.

**One correction to the framing**, since it changes who inherited this rather
than whether it needs fixing. The deleted Drizzle runner was *not* the migration
step for this database: `dev-local.ts` pointed it at `docker-compose.dev.yml`'s
separate `nuchi_dev` database on port **54329**, with its own credentials, which
the Go API never connected to. The Go API's database — `nuchi` on 5432 — has
never had a migration step in the root flow. So the gap predates this PR.

It is still exactly right to close it here. This PR is what rewrites the dev
flow and asserts that goose owns migrations, so shipping that assertion without
a step that runs goose would be the actual regression.

### 2. [P2] AGENTS.md internally inconsistent

Correct, and I had left it half-swept: the Env section still said to configure
`NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY` and `CLERK_SECRET_KEY` three lines below a
Stack section saying Clerk was gone.

Replaced with what is actually read, annotated by *who* reads it — the
distinction that matters for an agent:

- `AUTH_JWT_SECRET` — required, no default, `openssl rand -base64 48`, never committed
- `AUTH_COOKIE_SECURE` — `false` locally, must be `true` anywhere deployed
- `DATABASE_URL` — the Go API's, with a note that the frontend does not connect to Postgres
- `APP_BASE_URL`, `SMTP_ADDR`, `MAIL_FROM`
- `GO_API_URL` — the only one Next itself reads

It also now states that the frontend build needs **no** environment variables,
which is newly true as of this PR and is the thing most likely to be
misremembered.

While in that section I fixed two links pointing at absolute paths on a machine
that is not this one (`/home/gonzalo/projects/nuchi/...`) — one of them was the
`.env.example` reference inside the block being rewritten.

## Verification

Cold run — `tsconfig.tsbuildinfo` removed first:

| check | result |
| --- | --- |
| `bunx tsc --noEmit` | exit 0 |
| `bun run lint` | exit 0 |
| `bun test` | 193 pass, 8 skip, 0 fail — 201 across 20 files |
| `bun run build` | passes with no Clerk keys in the environment |

Also re-ran the parent-`node_modules` fallthrough check that caught the earlier
`@types/pg` failure. `pg` now resolves from the worktree's own tree; the only
remaining hit is `server-only`, which is not in `package.json` or the lockfile
and is probed only by local `.next/` artifacts that CI does not have when it
typechecks.

scripts/dev-local.ts
GonzaloSecades: Fixed in `84ef2b0`, and verified against source first — every part of this holds.

`01-app-role.sql` creates the role, the schema grant and `citext`, and no tables. `main.go` has no migration call, and on `VerifyRLSActive` failure it logs, closes the pool and `os.Exit(1)` (lines 44-48). `ownedTablesRequiringRLS` is exactly `{accounts, categories, transactions}`.

The failure is also more misleading than "RLS is not active": `row_security_active($1::text)` **errors** on a table that does not exist, so a clean checkout got a startup error naming row level security when the real problem was an empty schema.

Added the explicit goose step to the canonical "Running Next and Go Together" flow, between Postgres and the API, with the pinned install command and why it is not optional. This file now points at it too, so removing the Drizzle runner does not read as "migrations do not matter".

I chose that over invoking goose here: it would put the backend's schema ownership and a Go toolchain dependency inside the frontend dev server, which is the coupling this PR removes.

One correction to the framing — it changes who inherited this, not whether to fix it. The deleted Drizzle runner was not the migration step for this database: it ran against `docker-compose.dev.yml`'s separate `nuchi_dev` on port 54329, which the Go API never connected to. The Go API's database has never had a migration step in the root flow, so the gap predates this PR. Still the right place to close it, since this PR is what asserts goose owns migrations.

AGENTS.md
GonzaloSecades: Fixed in `84ef2b0`. You are right, and I had left the sweep half-done — the Env section told agents to configure Clerk keys three lines below a Stack section saying Clerk was gone.

Replaced with what is actually read, annotated by *who* reads it, since that is the distinction that matters for an agent:

- `AUTH_JWT_SECRET` — required, no default, `openssl rand -base64 48`, never committed
- `AUTH_COOKIE_SECURE` — `false` locally, must be `true` anywhere deployed
- `DATABASE_URL` — the Go API's, with a note that the frontend does not connect to Postgres
- `APP_BASE_URL`, `SMTP_ADDR`, `MAIL_FROM`
- `GO_API_URL` — the only one Next itself reads

It also now states that the frontend build needs **no** environment variables, which is newly true as of this PR and is the thing most likely to be misremembered.

While in that block I also fixed two links pointing at absolute paths on a machine that is not this one (`/home/gonzalo/projects/nuchi/...`); one of them was the `.env.example` reference inside the section being rewritten.

scripts/dev-local.ts
GonzaloSecades: Follow-up [P2]: the new migration step still does not bootstrap a fresh shell as written. The setup section tells the developer to copy values into `.env.local`, but neither goose nor the Go process loads that file: `backend/internal/config/config.go` reads only `os.Getenv`. Consequently `"$DATABASE_URL"` is empty here; goose falls back to libpq's current-user defaults (I reproduced it attempting `user=zecan database=`) and fails authentication. The following `go run ./cmd/api` also lacks the required exported `AUTH_JWT_SECRET`. Please make this canonical flow self-contained—e.g. explicitly export/source the required values, or pass the documented local database URL to goose and export a generated JWT secret before starting the API—so the fresh-checkout sequence actually runs.

AGENTS.md
GonzaloSecades: Follow-up [P2]: the Clerk entries are gone, but the replacement statement is still inaccurate. `lib/api-base-url.ts:8` reads `process.env.NEXT_PUBLIC_API_URL` on the server, and this PR deliberately retains that variable in `.env.example` and the CI comments. Therefore `GO_API_URL` is not the only setting read by Next/TypeScript. Please add `NEXT_PUBLIC_API_URL` back as the optional server-side API-base setting (or narrow the introductory claim) so the authoritative guide matches the surviving code.

scripts/dev-local.ts
GonzaloSecades: Fixed in `f9b0849`, and I reproduced your failure rather than taking it on trust. Fresh shell, `DATABASE_URL` unset:

```
goose run: failed to ensure DB version: failed to connect to `user=zecan database=`:
failed SASL auth: FATAL: password authentication failed for user "zecan" (SQLSTATE 28P01)
```

Same `user=zecan database=` you saw. Confirmed the cause too: there is no godotenv anywhere in `backend/`, and `config.go` reads `os.Getenv` only, so `.env.local` never reaches the Go side.

The goose step now spells the connection string out literally, with the reason it does not use `"$DATABASE_URL"`. Verified against a live Postgres:

```
goose -dir migrations postgres 'postgres://nuchi:nuchi@localhost:5432/nuchi?sslmode=disable' status
-> all five migrations listed as applied, exit 0
```

The API step now exports `AUTH_JWT_SECRET`. I checked which values actually need exporting rather than listing them all: `DATABASE_URL`, `SMTP_ADDR`, `MAIL_FROM` and `APP_BASE_URL` all have local defaults in `config.go`, and `AUTH_JWT_SECRET` is the only one with none — so it is the single export in the flow. That also matches `backend/README.md`'s own Local Run section, which had it right all along; the root README had diverged from it.

I also fixed what I think is the misconception underneath this: the Environment section told you to copy backend variables into `.env.local`, which the backend will never read. It now says so.

**Noticed, not changed:** `backend/README.md`'s migrations table lists only `00001`-`00003`, but `goose status` shows five — `00004_transactions_category_owner_rls.sql` and `00005_transactions_amount_bigint.sql` are missing from it. `backend/**` is outside this PR's scope, so I have left it; worth a follow-up.

AGENTS.md
GonzaloSecades: Fixed in `f9b0849`. You are right — `lib/api-base-url.ts:8` reads `process.env.NEXT_PUBLIC_API_URL` on the server, and this PR deliberately keeps that variable, so "the only one Next itself reads" was simply false.

The section is now split by which process reads what, which is the distinction that was missing:

**Read by Next** — `GO_API_URL`, and `NEXT_PUBLIC_API_URL` restored as the optional server-side API base, with the note that it is read only on the server and that leaving it unset is correct behind the proxy (in the browser the base is always relative).

**Read by the Go API** — `AUTH_JWT_SECRET`, `AUTH_COOKIE_SECURE`, `DATABASE_URL`, `APP_BASE_URL`, `SMTP_ADDR`, `MAIL_FROM`.

With the point that nothing bridges the two: `.env.local` is loaded by Bun for Next, and the Go API reads `os.Getenv` only.

I also tightened the opening claim while I was there — it asserted that nothing was required, two paragraphs above a required `AUTH_JWT_SECRET`. It now says the accurate and more useful thing: `bun run build` and `bun dev` need no environment variables, and `AUTH_JWT_SECRET` is the only value with no default and therefore the only one that must be exported for `go run ./cmd/api`.

scripts/dev-local.ts
GonzaloSecades: Good catch, and a fair one — I fixed the README and left the pointer to it broken. `scripts/dev-local.ts` still showed `"$DATABASE_URL"`, in the file someone is most likely to open when wondering where migrations went.

Corrected in `cb9421a`: literal URL, plus the reason it is literal — nothing exports `$DATABASE_URL`, so it expands to an empty string and libpq falls back to the OS username.

Swept for other instances. One remains: **`backend/README.md:327`** has the identical `goose -dir migrations postgres "$DATABASE_URL" up`, with nothing exporting the variable in its own Local Run section either. That file is outside this ticket's scope so I have not touched it — worth folding into a follow-up together with its migrations table, which lists `00001`-`00003` while `goose status` reports five.

PR #130, opened 2026-08-17
Add one-command local dev stack

GonzaloSecades:
## What changed

- Reworked `bun dev` into a full local stack launcher: Docker Postgres/Mailpit, Postgres readiness, goose migrations, Go API health check, then Next dev.
- Added `bun run dev:next` for frontend-only manual debugging.
- Moved the local Compose Postgres host port default to `54329` with `POSTGRES_PORT` override support, avoiding conflicts with a host Postgres already bound to `5432`.
- Updated `.env.example`, frontend/backend README docs, and the Go backend local default database URL to match the new dev setup.

## Why

After the Hono-to-Go migration, local development still required a multi-terminal sequence and manual migration/API startup. This PR makes the default dev path one command while preserving manual commands for debugging individual pieces.

## Validation

- `bunx tsc --noEmit`
- `bun run lint`
- `cd backend && go test ./internal/config`
- `bun dev` smoke test: started Compose services, migrated DB to version 5, Go API verified RLS and served on `8080`, Next became ready on `3000`.

## Notes

- `graphify update .` was run after the code changes, but `graphify-out/` output was intentionally left uncommitted per repo guidance.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,7 +1,7 @@
 # Local PostgreSQL connection string, read by the Go API. The frontend does not
 # connect to Postgres.
-# Docker Compose default: postgres://nuchi:nuchi@localhost:5432/nuchi?sslmode=disable
-DATABASE_URL=postgres://nuchi:nuchi@localhost:5432/nuchi?sslmode=disable
+# Docker Compose host default: 54329 -> container 5432
+DATABASE_URL=postgres://nuchi:nuchi@localhost:54329/nuchi?sslmode=disable
 
 # Public base URL used by the frontend API client.
 # Local Next.js default: http://localhost:3000
@@ -38,6 +38,7 @@ AUTH_COOKIE_SECURE=false
 # Local app/runtime flags.
 # APP_ENV is informational for local tooling.
 APP_ENV=local
+POSTGRES_PORT=54329
 
 # Go backend scaffold configuration.
 # Defaults from backend/README.md: 0.0.0.0:8080
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -13,16 +13,19 @@
 
 ## Commands
 
-- `bun dev`
+- `bun dev` — starts Docker Postgres/Mailpit, applies goose migrations, builds
+  and starts the Go API, waits for `/api/health`, then starts Next.
+- `bun run dev:next` — Next-only manual debugging when Docker, migrations, and
+  the Go API are already running.
 - `bun run lint`
 - `bun run build`
 
 ## Env
 
-`bun run build` and `bun dev` need no environment variables at all. The list
-splits by which process reads them, and **nothing bridges the two**:
-`.env.local` is loaded by Bun for the Next process, while the Go API reads
-`os.Getenv` only, so its values must be exported in the shell that starts it.
+`bun run build` needs no environment variables. `bun dev` reads `.env.local`
+through Bun, supplies local defaults, and passes that environment to the Go API
+child process. If you run `go run ./cmd/api` manually, export backend variables
+in that shell; the Go API itself still reads only `os.Getenv`.
 
 Read by **Next**:
 
@@ -31,24 +34,27 @@ Read by **Next**:
 - `NEXT_PUBLIC_API_URL` — optional origin for the API client. Read only on the
   server (`lib/api-base-url.ts`); in the browser the base is always relative so
   requests stay same-origin. Leaving it unset is correct behind the proxy.
+- `POSTGRES_PORT` — host port published by Docker Compose for local Postgres.
+  Defaults to `54329`.
 
 Read by the **Go API**:
 
-- `AUTH_JWT_SECRET` — required, no default. The API exits at startup if it is
-  unset or shorter than 32 bytes. Generate with `openssl rand -base64 48`;
-  never commit a value.
+- `AUTH_JWT_SECRET` — required by the Go API, no backend default. `bun dev`
+  generates an ephemeral local value when it is unset; manual API runs must
+  export one. Generate with `openssl rand -base64 48`; never commit a value.
 - `AUTH_COOKIE_SECURE` — `false` locally; must be `true` anywhere deployed,
   which requires HTTPS.
 - `DATABASE_URL` — the Go API's connection string. The frontend does not connect
-  to Postgres.
+  to Postgres. Local default:
+  `postgres://nuchi:nuchi@localhost:54329/nuchi?sslmode=disable`.
 - `APP_BASE_URL` — origin used to build links in verification and reset mail;
   origin-only and validated at backend startup.
 - `SMTP_ADDR`, `MAIL_FROM` — outgoing mail. `SMTP_ADDR` is one `host:port`.
 
-`AUTH_JWT_SECRET` is the only one with no default, so it is the only one that
-must be exported for `go run ./cmd/api`. Reference:
-[`.env.example`](.env.example), and the full backend table in
-[`backend/README.md`](backend/README.md).
+`bun dev` fails fast if local `DATABASE_URL` and `POSTGRES_PORT` disagree, to
+avoid running migrations against a stale `localhost:5432` database from an old
+`.env.local`. Reference: [`.env.example`](.env.example), and the full backend
+table in [`backend/README.md`](backend/README.md).
 
 ## Repo Rules
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -34,8 +34,16 @@ Read by **Next**:
 - `NEXT_PUBLIC_API_URL` — optional origin for the API client. Read only on the
   server (`lib/api-base-url.ts`); in the browser the base is always relative so
   requests stay same-origin. Leaving it unset is correct behind the proxy.
-- `POSTGRES_PORT` — host port published by Docker Compose for local Postgres.
-  Defaults to `54329`.
+
+Read by **`bun dev` and Docker Compose** (neither Next nor the Go API reads
+these):
+
+- `POSTGRES_PORT` — host port Compose publishes for local Postgres. Defaults to
+  `54329`. `docker compose` interpolates it from `.env`, while `bun dev` picks
+  it up from `.env.local`.
+- `COMPOSE_PROJECT_NAME` — Compose project the local services belong to.
+  `bun dev` defaults it to `nuchi` so a worktree reuses the same containers
+  instead of standing up a second set named after its directory.
 
 Read by the **Go API**:
 
@@ -51,10 +59,13 @@ Read by the **Go API**:
   origin-only and validated at backend startup.
 - `SMTP_ADDR`, `MAIL_FROM` — outgoing mail. `SMTP_ADDR` is one `host:port`.
 
-`bun dev` fails fast if local `DATABASE_URL` and `POSTGRES_PORT` disagree, to
-avoid running migrations against a stale `localhost:5432` database from an old
-`.env.local`. Reference: [`.env.example`](.env.example), and the full backend
-table in [`backend/README.md`](backend/README.md).
+`bun dev` runs `goose up` unattended, so it checks `DATABASE_URL` before doing
+anything: it refuses a non-local host outright, and fails fast when a local host
+disagrees with `POSTGRES_PORT`. Both cases come from a stale `.env.local` — a
+Neon URL left over from before #85, or `localhost:5432` from before the Compose
+port moved — and both would otherwise migrate a database that never asked for
+it. Reference: [`.env.example`](.env.example), and the full backend table in
+[`backend/README.md`](backend/README.md).
 
 ## Repo Rules
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,7 +12,7 @@ services:
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD: postgres
     ports:
-      - "5432:5432"
+      - '${POSTGRES_PORT:-54329}:5432'
     volumes:
       - nuchi-postgres-data:/var/lib/postgresql/data
       # Init scripts run only on first initialization of an empty data
@@ -21,7 +21,7 @@ services:
       # `docker compose down --volumes` to pick this up.
       - ./docker/postgres/init:/docker-entrypoint-initdb.d:ro
     healthcheck:
-      test: ["CMD-SHELL", "pg_isready -U postgres -d nuchi"]
+      test: ['CMD-SHELL', 'pg_isready -U postgres -d nuchi']
       interval: 2s
       timeout: 5s
       retries: 20
@@ -29,8 +29,8 @@ services:
   mailpit:
     image: axllent/mailpit:latest
     ports:
-      - "1025:1025"
-      - "8025:8025"
+      - '1025:1025'
+      - '8025:8025'
 
 volumes:
   nuchi-postgres-data:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,7 +21,7 @@ services:
       # `docker compose down --volumes` to pick this up.
       - ./docker/postgres/init:/docker-entrypoint-initdb.d:ro
     healthcheck:
-      test: ['CMD-SHELL', 'pg_isready -U postgres -d nuchi']
+      test: ["CMD-SHELL", "pg_isready -U postgres -d nuchi"]
       interval: 2s
       timeout: 5s
       retries: 20
@@ -29,8 +29,8 @@ services:
   mailpit:
     image: axllent/mailpit:latest
     ports:
-      - '1025:1025'
-      - '8025:8025'
+      - "1025:1025"
+      - "8025:8025"
 
 volumes:
   nuchi-postgres-data:
--- a/scripts/dev-local.ts
+++ b/scripts/dev-local.ts
@@ -1,6 +1,7 @@
 import { randomBytes } from 'node:crypto';
 import { spawn } from 'node:child_process';
-import { existsSync } from 'node:fs';
+import { existsSync, unlinkSync } from 'node:fs';
+import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 
 /**
@@ -20,6 +21,8 @@ const HEALTH_RETRIES = 30;
 const HEALTH_DELAY_MS = 1000;
 const API_HEALTH_RETRIES = 30;
 const API_HEALTH_DELAY_MS = 1000;
+// Keep this aligned with .github/workflows/ci.yml's installed goose version.
+const GOOSE_VERSION = 'v3.27.2';
 const POSTGRES_PORT = process.env.POSTGRES_PORT ?? '54329';
 const DATABASE_URL =
   process.env.DATABASE_URL ??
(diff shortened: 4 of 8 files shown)


scripts/dev-local.ts
```diff
@@ -1,60 +1,107 @@
...
  */
 const COMPOSE_SERVICES = ['postgres', 'mailpit'] as const;
 const HEALTH_RETRIES = 30;
 const HEALTH_DELAY_MS = 1000;
+const API_HEALTH_RETRIES = 30;
+const API_HEALTH_DELAY_MS = 1000;
+const POSTGRES_PORT = process.env.POSTGRES_PORT ?? '54329';
+const DATABASE_URL =
+  process.env.DATABASE_URL ??
+  `postgres://nuchi:nuchi@localhost:${POSTGRES_PORT}/nuchi?sslmode=disable`;
```
GonzaloSecades, line 29: **This is the one I would fix before merge.**

`process.env.DATABASE_URL` wins here, and Bun loads `.env.local` into this script's environment. Anyone who set up from the previous `.env.example` has `DATABASE_URL=postgres://nuchi:nuchi@localhost:5432/nuchi?sslmode=disable` in that file, so after this PR:

1. Compose publishes Postgres on `54329`.
2. `waitForPostgres()` passes — it runs `pg_isready` via `docker compose exec`, *inside* the container, so it never validates the published host port.
3. goose then connects to `localhost:5432` — the host Postgres this PR exists to avoid colliding with — and applies `backend/migrations/` there.

Best case it fails with a confusing connection error. Worst case the developer has a `nuchi` role/database on their host instance and the migrations land in it.

`POSTGRES_PORT` and `DATABASE_URL` are two independent knobs for one fact, so the same mismatch is reachable even on a fresh setup (`POSTGRES_PORT=6000` with the example `DATABASE_URL` left alone).

Suggestions, roughly in order of preference:

- Parse `DATABASE_URL` and fail fast when its port disagrees with `POSTGRES_PORT`, naming both values — it catches the stale-`.env.local` case and the drift case with one check.
- Or `log()` the resolved `DATABASE_URL` (host/port at least) before the goose step, so the target is visible rather than inferred.
- Either way, add an upgrade note to the README: existing `.env.local` files need `5432` → `54329`. This PR changes a default that people already have written down.

GonzaloSecades (commented):
## Review: one-command local dev stack

The direction is right and the docs rewrite is genuinely good — the manual flow is preserved rather than deleted, and the README now reads as "here is the one command, here is what it does for you, here is how to do it by hand." CI is green on all three jobs, and the port move away from `5432` is a sensible fix for the host-Postgres collision.

One thing I think should be fixed before merge, three worth a look, and some polish. Details inline; summary here.

### Should fix

**A stale `.env.local` silently wins over the new Compose port — and can migrate the wrong database.** `DATABASE_URL` is resolved as `process.env.DATABASE_URL ?? …${POSTGRES_PORT}…`, and Bun loads `.env.local` into this script. Every developer who copied the old `.env.example` has `DATABASE_URL=…@localhost:5432/nuchi…` sitting there. After this PR, Compose publishes on `54329`, the readiness probe passes (it runs `pg_isready` *inside* the container, so it never touches the host port), and then goose runs against whatever is on host `5432`. In the exact scenario this PR is motivated by — a developer with a host Postgres bound to `5432` — that means applying `backend/migrations/` to their unrelated local database. See the inline comment for suggested fixes; at minimum this needs an upgrade note in the README.

### Worth a look

- `waitForApi()` is not raced against the API process exiting, so a Go compile error becomes 30 s of silence followed by the wrong error message.
- `proc.kill()` on `go run` does not stop the compiled binary. I verified this locally: SIGTERM to the `go run` parent, and the child kept writing to stdout for the remaining 3 s of the test. Ctrl-C in a terminal is fine (the whole foreground process group gets SIGINT), but the `main().catch()` path orphans the API holding `:8080` — and the next `bun dev` will then get a healthy `/api/health` from the *stale* binary.
- **`AGENTS.md` was not updated and is now wrong.** It still says `bun dev` needs "no environment variables at all" and that **"nothing bridges the two"** — `.env.local` for Next, `os.Getenv` for Go. Bridging those two is precisely what this PR does: `baseEnv` is built from Bun-loaded `.env.local` and handed to the Go child. `README.md` and `backend/README.md` were updated carefully; the Codex guide needs the same treatment, since it is the file the other agent reads.

### Polish

- The `existsSync(nextBin)` preflight runs *after* Docker, migrations and the API have started. It costs nothing to check first.
- `prefixLines()` splits on chunk boundaries rather than line boundaries, so a `[next] ` prefix can land mid-line on a split write. Also, piping Next's output loses its TTY, so you lose colors and its interactive output.
- The old `process.exit(code ?? 0)` is gone, so the script no longer propagates the dev server's exit code.
- `v3.27.2` is now pinned in four places (this script, both READMEs, `ci.yml`). Worth a comment pointing at `ci.yml` as the source of truth so they do not drift.
- The script injects `NEXT_PUBLIC_API_URL`, but nothing in the repo reads it any more — `grep` across `*.ts`/`*.tsx` returns nothing. Pre-existing debt, not introduced here, but this PR is a good moment to drop it.
- README: "leaves Docker volumes running" should be *containers* — volumes do not run.

### Verified, for the record

- All three CI jobs pass; the `defaultDatabaseURL` change is safe for CI, which sets `TEST_DATABASE_URL`/`ADMIN_DATABASE_URL` explicitly rather than relying on the Go default.
- `/api/health` is the correct path (`backend/internal/http/router.go:29`), and `GO_API_URL=http://localhost:8080` is origin-only, so it passes `normalizeGoApiUrl`.
- Refresh tokens are opaque and stored hashed (`refresh_tokens.token_hash`), so the regenerated `AUTH_JWT_SECRET` invalidates access tokens only, not sessions. Milder than I first assumed — hence polish, not a blocker.
- Cold-cache build of `./cmd/api` here is ~12.5 s, comfortably inside the 30 s health budget on this machine. On a slow laptop with a cold module cache it could get tight, which is another reason to surface the process exit rather than only timing out.

scripts/dev-local.ts
```diff
@@ -112,36 +155,108 @@ async function waitForPostgres() {
...
+      'up',
+    ],
+    { cwd: 'backend' }
+  );
+
+  log('Starting Go API');
+  const api = spawnLogged(['go', 'run', './cmd/api'], 'api', {
+    cwd: 'backend',
+  });
+  await waitForApi();
```
GonzaloSecades: `waitForApi()` only polls the health endpoint — it never observes `api`. `watchProcess(api, …)` is not attached until line 247, after this await returns.

So when `go run ./cmd/api` fails to compile, or exits on a config error (`AUTH_JWT_SECRET` too short, `VerifyRLSActive` refusing an unmigrated database), you get 30 s of nothing followed by `Go API did not become healthy at …` — while the actual Go error already scrolled past under the `[api]` prefix. That is the failure mode a first-time setup is most likely to hit.

Racing the poll against the process exit fixes both the latency and the message:

```ts
await Promise.race([
  waitForApi(),
  new Promise<never>((_, reject) => {
    api.once('exit', (code) =>
      reject(new Error(`Go API exited with code ${code} before becoming healthy`))
    );
  }),
]);
```

That also removes the pressure on the 30 s budget for a cold `go build`.

scripts/dev-local.ts
```diff
@@ -112,36 +155,108 @@ async function waitForPostgres() {
...
+      'migrations',
+      'postgres',
+      DATABASE_URL,
+      'up',
+    ],
+    { cwd: 'backend' }
+  );
+
+  log('Starting Go API');
+  const api = spawnLogged(['go', 'run', './cmd/api'], 'api', {
```
GonzaloSecades: `stopChildren()` calls `child.kill()`, which sends SIGTERM to the `go run` process — not to the binary `go run` compiled and exec'd. I checked this rather than assuming: SIGTERM to a `go run` parent, and the child kept writing to stdout for the remaining 3 s of the test.

Ctrl-C in a terminal is fine, because SIGINT goes to the entire foreground process group and the compiled binary receives it directly. The paths that break are the programmatic ones — `main().catch()` → `stopChildren()` → `process.exit(1)`, or a SIGTERM from a supervisor. There the API binary is orphaned still holding `:8080`, and the API only logs on startup and per request, so it will not die of SIGPIPE any time soon.

The nasty part is the next run: `waitForApi()` gets a `200` from the orphaned old binary and the script continues happily, while the new `go run` fails to bind. You end up debugging a dev server that is serving the previous build.

Two ways out:

- `spawn(..., { detached: true })` and kill the group with `process.kill(-proc.pid, 'SIGTERM')`.
- Or drop `go run` for the API: `go build -o ./tmp/api ./cmd/api` then spawn that binary directly. One process instead of two, `kill()` means what it says, the compile error surfaces as a failed `run()` step before anything is spawned, and restarts get faster.

scripts/dev-local.ts
```diff
@@ -1,60 +1,107 @@
...
+  `http://localhost:${process.env.BACKEND_PORT ?? '8080'}`;
 
 const baseEnv = {
   ...process.env,
-  APP_ENV: 'local',
+  APP_ENV: process.env.APP_ENV ?? 'local',
+  APP_BASE_URL: process.env.APP_BASE_URL ?? 'http://localhost:3000',
+  AUTH_COOKIE_SECURE: process.env.AUTH_COOKIE_SECURE ?? 'false',
+  AUTH_JWT_SECRET:
+    process.env.AUTH_JWT_SECRET ?? randomBytes(48).toString('base64'),
```
GonzaloSecades: This regenerates a signing secret on every `bun dev`, silently.

The blast radius is smaller than it looks — refresh tokens are opaque and stored hashed (`refresh_tokens.token_hash`), so sessions survive a restart and only previously-minted access tokens stop verifying. But it is invisible: a developer restarting the stack sees a 401 blip with nothing in the logs explaining why, and any token they had copied for a manual `curl` is silently dead.

A `log()` line when the secret is generated ("generated an ephemeral AUTH_JWT_SECRET; set one in .env.local to keep access tokens valid across restarts") would cover it. Persisting a generated value into `.env.local` would be friendlier still, but writing to that file from a dev script is probably more magic than this PR wants.

scripts/dev-local.ts
```diff
@@ -112,36 +155,108 @@ async function waitForPostgres() {
...
-      ? './node_modules/.bin/next.cmd'
-      : './node_modules/.bin/next';
+      ? join('.', 'node_modules', '.bin', 'next.cmd')
+      : join('.', 'node_modules', '.bin', 'next');
 
-  const next = spawn(nextBin, ['dev'], {
-    env: baseEnv,
-    stdio: 'inherit',
-  });
+  if (!existsSync(nextBin)) {
```
GonzaloSecades: This preflight fires only after Docker is up, migrations have run and the API is healthy — so the "run `bun install` first" case costs the developer the whole startup sequence before it tells them.

Move the `nextBin` resolution and this check to the top of `main()`, before the Compose step. Same message, no side effects first.

docker-compose.yml
```diff
@@ -21,16 +21,16 @@ services:
       # `docker compose down --volumes` to pick this up.
       - ./docker/postgres/init:/docker-entrypoint-initdb.d:ro
     healthcheck:
-      test: ["CMD-SHELL", "pg_isready -U postgres -d nuchi"]
+      test: ['CMD-SHELL', 'pg_isready -U postgres -d nuchi']
```
GonzaloSecades: Unrelated churn: this line and the two `mailpit` port lines only changed quote style. The `POSTGRES_PORT` interpolation on line 15 needs quoting; these three did not change meaning. Worth reverting to keep the diff to the actual change.

.env.example
```diff
@@ -38,6 +38,7 @@ AUTH_COOKIE_SECURE=false
 # Local app/runtime flags.
 # APP_ENV is informational for local tooling.
 APP_ENV=local
+POSTGRES_PORT=54329
```
GonzaloSecades, line 41: Two things worth noting here.

First, this and the `DATABASE_URL` on line 4 both encode the port, independently — see the comment on `scripts/dev-local.ts` about what happens when they disagree.

Second, Compose interpolates `${POSTGRES_PORT}` from `.env` in the project directory, **not** from `.env.local`. So setting `POSTGRES_PORT` here does nothing for the manual `docker compose up -d postgres` flow the README documents — that path always gets the `54329` fallback, while `bun dev` honors the override because the script passes it through `baseEnv`. Fine as long as nobody overrides it; worth a one-line note in the README's Docker section so the discrepancy is not discovered the hard way.

scripts/dev-local.ts
```diff
@@ -68,11 +101,41 @@ async function run(
...
+  let parsed: URL;
+  try {
+    parsed = new URL(DATABASE_URL);
+  } catch {
+    throw new Error(
+      `DATABASE_URL must be an absolute Postgres URL, got: ${DATABASE_URL}`
+    );
+  }
+
+  const localHosts = new Set(['localhost', '127.0.0.1', '::1']);
```
GonzaloSecades: `'::1'` in this set can never match. `URL.hostname` returns IPv6 hosts **with** the brackets, so a `DATABASE_URL` of `postgres://nuchi:nuchi@[::1]:5432/nuchi` yields `[::1]`, falls through the `has()` check, and returns early — skipping the very validation this function exists for.

Verified:

```
postgres://…@localhost:54329/… → hostname "localhost"  matches: true
postgres://…@127.0.0.1:5432/…  → hostname "127.0.0.1"  matches: true
postgres://…@[::1]:5432/…      → hostname "[::1]"      matches: false
```

Adding `'[::1]'` to the set covers it:

```ts
const localHosts = new Set(['localhost', '127.0.0.1', '::1', '[::1]']);
```

Narrow in practice — nobody is likely to write the loopback that way in `.env.local` — but the set already states the intent to cover it, so the entry should work.

While you are here: the message on line 123 hardcodes `localhost:${databasePort}` even when the host is `127.0.0.1`. Interpolating `parsed.hostname` keeps the error honest about what it actually read.

scripts/dev-local.ts
```diff
@@ -112,36 +176,158 @@ async function waitForPostgres() {
...
+
+  log('Starting Go API');
+  const api = spawnLogged([API_BIN]);
+  await waitForApi(api);
+
+  log('Starting Next dev server');
+  const next = spawnLogged([nextBin, 'dev']);
 
-  process.exit(code ?? 0);
+  process.on('SIGINT', stopChildren);
```
GonzaloSecades: Nit: these register after the API is spawned on line 300, so a Ctrl-C during `waitForApi` — up to 30 s — kills the script with no handler installed, `.finally` never runs, and the temp binary in `tmpdir()` is left behind. The API process itself is fine (same foreground process group, so it gets the SIGINT directly); it is only the cleanup that is skipped.

Moving both `process.on` registrations above the `go build` step would make the cleanup path unconditional.

GonzaloSecades (commented):
## Follow-up review of a4cc5ed

Every item from the last round is addressed, and two of them came back better than what I suggested. CI is green on all three jobs.

### First — a correction on my part

I said `NEXT_PUBLIC_API_URL` was dead code and nothing read it. **That was wrong**, and it was the basis for you dropping it from `baseEnv`. It is read at `lib/api-base-url.ts:8`. My grep had an unquoted `--include=*.ts` that the shell expanded before `grep` saw it, so it matched nothing and I reported that as an absence.

The removal is still fine, for a different reason than the one I gave: `resolveApiBaseUrl()` only consults the variable on the server, and every call site of `apiClient` today reaches it through client components (`lib/auth/session.tsx`, the two `(auth)` pages), so the browser path returns `''` and stays same-origin regardless. Unset really is correct behind the proxy, exactly as `AGENTS.md` says — and not injecting a default now makes `bun dev` match that documented posture instead of quietly overriding it. `.env.example` still ships the value, so nothing changes for anyone set up from it either.

No action needed. But the variable is live, so it should not get pruned from `.env.example` or the docs on the strength of my earlier claim.

### Fixes I checked

- **Port mismatch** — `validateDatabaseUrlPort()` runs first thing in `main()`, before Compose, with an error naming both values and the expected URL. Plus `databaseTarget()` in the migration log line, and upgrade notes in both READMEs. This is the fix; it closes the wrong-database path.
- **`waitForApi` race** — correctly done: the `settled` guard is checked inside the poll loop, and `api.off('exit', onExit)` removes the listener on every settle path, so no stray handler survives into the `watchProcess` phase.
- **`go run` orphan** — you replaced it with `go build -o` plus a direct exec of the binary, which is the better of the two options I offered. `kill()` now reaches the actual process, the compile error surfaces as a failed `run()` step before anything is spawned, and the 30 s health budget no longer has to cover compilation. Temp path is pid-scoped and `.exe`-suffixed on Windows.
- **`AGENTS.md`** — the "nothing bridges the two" claim is gone and the Env section now describes what actually happens. One bucket nit inline.
- Preflight order, `prefixLines`, exit code via `process.exitCode`, `GOOSE_VERSION` with the pointer to `ci.yml`, the reverted quote churn, "containers" not "volumes", and the `.env` vs `.env.local` Compose note — all done.

On dropping the `[api]`/`[next]` prefixes with `stdio: 'inherit'`: that trades per-process attribution for Next keeping its TTY. I think it is the right call — Go's slog lines are self-identifying enough in practice — just noting it was a real trade and not a free win.

### Small things left

One actual defect in the new guard (the `::1` entry never matches) and two nits, all inline. None of them block.

AGENTS.md
```diff
@@ -31,24 +34,27 @@ Read by **Next**:
 - `NEXT_PUBLIC_API_URL` — optional origin for the API client. Read only on the
   server (`lib/api-base-url.ts`); in the browser the base is always relative so
   requests stay same-origin. Leaving it unset is correct behind the proxy.
+- `POSTGRES_PORT` — host port published by Docker Compose for local Postgres.
```
GonzaloSecades: `POSTGRES_PORT` is filed under **Read by Next**, but Next never reads it — the dev script reads it, and Docker Compose interpolates it into the port mapping. This section's whole organizing idea is which process reads what, so it is the one place the misfiling costs something.

A third bucket, or a line under Commands, would fit better — something like "Read by `bun dev` / Docker Compose: `POSTGRES_PORT`, `COMPOSE_PROJECT_NAME`", since `COMPOSE_PROJECT_NAME` is now an honored override too and is not documented anywhere in this file.

GonzaloSecades:
## The port guard skipped remote hosts — and it fired for real

Follow-up to the review above, pushed as 67e30f0. Recording this because it is the reason the guard changed shape, and because it is a live hazard for anyone whose `.env.local` predates #85.

### What happened

While A/B testing the interrupt-cleanup fix, I ran `bun dev` on the pre-fix code without first checking what the local `.env.local` pointed at. It did not point at localhost — it still held the legacy Neon `DATABASE_URL` from before #85.

`validateDatabaseUrlPort()` checked the port only when the hostname was local, and **returned early for everything else**:

```ts
const localHosts = new Set(['localhost', '127.0.0.1', '::1']);
if (!localHosts.has(parsed.hostname)) {
  return;          // ← a remote host skipped validation entirely
}
```

So the guard did nothing, and goose ran:

```
[dev] Applying backend migrations to ep-dark-river-…neon.tech:5432/neondb
2026/08/17 00:03:46 OK   00001_auth_base.sql (511.35ms)
2026/08/17 00:03:46 goose run: ERROR 00002_finance_base.sql: failed to run SQL migration:
  … CREATE TABLE accounts … : ERROR: relation "accounts" already exists (SQLSTATE 42P07)
```

`00001_auth_base.sql` applied to the remote database — `goose_db_version` stamped at version 1, plus `users`, `email_verification_tokens`, `password_reset_tokens`, `refresh_tokens`, three indexes and `CREATE EXTENSION IF NOT EXISTS citext`. `00002` aborted on the pre-existing `accounts` table, which is the only reason it stopped there. Everything applied was `CREATE`; nothing was dropped, altered or overwritten.

### Why this is worth more than a nit

The original review framed the risk as "migrations land on a host Postgres on 5432." The remote case is strictly worse and was completely unguarded:

- It is **silent**. A local-port mismatch usually fails to connect. A stale Neon URL connects fine and migrates.
- It is **unattended**. `bun dev` runs `goose up` with no confirmation, as step three of a command whose advertised job is "start my local dev server."
- The stale value is **the default state** for anyone who set this repo up before #85 and never revisited `.env.local`. Nothing in the migration retired those URLs from developer machines.

### The change

`bun dev` provisions local Docker Postgres, so a non-local target is always a mistake here. It is now refused outright rather than skipped, with no override — migrate a remote database by hand, deliberately:

```
DATABASE_URL points at the remote host ep-dark-river-…neon.tech, but bun dev only drives the local Docker Postgres.
It would have applied backend/migrations/ to that database. Refusing.
Check DATABASE_URL in .env.local — a Neon URL from before #85 is the usual cause.
Expected local default: postgres://nuchi:nuchi@localhost:54329/nuchi?sslmode=disable
```

Verified against the exact URL that fired: exit 1, before Docker Compose or goose is invoked. The local-port mismatch path is unchanged apart from the `[::1]` fix.

Also in 67e30f0, from the review comments above: the bracketed IPv6 loopback (`URL.hostname` keeps the brackets, so `'::1'` could never match), the hostname in the mismatch message, the signal handlers moved above `go build` so an interrupt during the health wait still runs the temp-binary cleanup, and the `AGENTS.md` env buckets.

### Worth considering separately

A guard in the dev script protects the dev script. `goose up` against a remote database stays one stale shell variable away in the manual flow the READMEs document, and `backend/migrations/` is not something you want applied by accident anywhere. That is out of scope here, but it looks like a candidate for the improvement registry rather than something to leave implicit.

